PR #4, opened 2020-09-30
setup GitHub Action

skyzh:
In this PR, I set up basic CI for this project.

* fix several compile error
* update protobuf generated files with latest version of protoc and rust-protobuf
* add GitHub Action for test and benchmark
* add cargo fmt check, and format all codes
* add `Cargo.lock` into repo
* comment out `table` as it cannot compile for now

For some tests and benchmarks not passing on GitHub Action:
* `test_one_key` will timeout on CI, so I increased timeout for this test. (but it still fails randomly)
* `skiplist_write` will overflow `thread 'thread '<unnamed>' panicked at 'assertion failed: offset as usize + size <= self.core.cap', skiplist/src/arena.rs:50:9` still investigating
* these fixes will be sent as separate PRs

Running result of the workflows can be found at https://github.com/skyzh/agatedb/pull/1 . Workflow files will take effect in this repo after being merged.


.gitignore
```diff
@@ -2,10 +2,6 @@
 # will have compiled files and executables
 target/
 
-# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
```
BusyJay, line 5: Why? It's expected to ignore lock.

proto/src/meta.rs
```diff
@@ -1,11 +1,12 @@
-// This file is generated by rust-protobuf 2.10.1. Do not edit
+// This file is generated by rust-protobuf 2.17.0. Do not edit
```
BusyJay, line 1: How about leaving it untouched and switching to prost in next PR instead?

.gitignore
skyzh: Current GitHub Action workflow requires hashing `Cargo.lock` for caching. If this is intended, I'll configure it to cache for specific task instead of `Cargo.lock`.

proto/src/meta.rs
skyzh: Alright

BusyJay:
Seems clippy is not run.

skyzh:
> Seems clippy is not run.

Yes, as it will produce a lot of errors, I'll open a separate PR to fix clippy warnings and enable clippy.

BusyJay (approved):

PR #11, opened 2020-10-09
fix test and bench panic

skyzh:

**Benchmark failure**

This is done by simply changing the capacity of skiplist while running this unit test. The long term solution will be changing iteration number of Criterion.rs in each bench case.

**`test_one_key` failure**

From GitHub Action log I found that the 100th(/200) message timed out. I guess this is because that one pool (randomly) never spawns thread. So I forced thread pool to shutdown in this test, to make sure all threads are terminated before we check for the response.

However, even when I shutdown the pool, there will still be possibility of not receiving any message.

```
thread 'test_one_key' panicked at 'timeout on receiving r0 w0 msg', skiplist/tests/tests.rs:157:23
```

And then I guess this is caused by pools with the same prefix name, leading to threads of the same name. I changed prefix name of two pools to different ones `read` and `write`. This doesn't work.

Finally I merged two pools into one. And this works.

This bug won't affect other part of agatedb, as currently, yatp is only used in tests.


BusyJay:
That's strange. /cc @sticnarf

BusyJay:
>  I guess this is because that one pool (randomly) never spawns thread. 

`build()` returns after the threads are spawned.

BusyJay (commented):
The purpose of two pool is to make sure get are being executed along side with put. When putting them in one queue, they are nearly always being executed one by one.

sticnarf:
I cannot reproduce the failure of `test_one_key` with master (https://github.com/tikv/agatedb/commit/fc497f1c00423180f9a448b92a581cf43fcd3458). Is there any special requirement to reproduce?

skyzh:
> I cannot reproduce the failure of `test_one_key` with master ([fc497f1](https://github.com/tikv/agatedb/commit/fc497f1c00423180f9a448b92a581cf43fcd3458)). Is there any special requirement to reproduce?

This only occurs on GitHub Action. You may send a PR to trigger GitHub Action build.

skyzh:
@sticnarf has verified that the test failure is related to rust toolchain version. I'll revert some changes in this PR and fix rust toolchain to another version.

sticnarf:
Sorry after some more attempts, I can produce this issue on different toolchains. I'll investigate the problem.

skyzh:
The case is that, even if `write_pool` spawns task before `read_pool` in program order, it is possible that `read_pool` gets executed first. Therefore, I made the following modifications:
* write an initial value to the key
* `yield` after each task

rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2020-07-01
```
BusyJay, line 1: We don't specify toolchain in a library project.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

PR #12, opened 2020-10-09
implement SST and iterator

skyzh:


BusyJay:
I suggest to use a minimal workspace. We need to publish it to crates.io one day, more packages means more maintenance.

skyzh:
> I suggest to use a minimal workspace. We need to publish it to crates.io one day, more packages means more maintenance.

Ok, I'll keep them in one crate.

PR #13, opened 2020-10-09
complete table builder

skyzh:
This PR fixes compile errors and some logical errors in current table builder. It is now in sync with current badger master.

I'll demonstrate related changes by commenting on code directly.

The CI failure will be resolved after https://github.com/tikv/agatedb/pull/11/ is ready.


proto/src/proto/meta.proto
```diff
@@ -55,7 +55,7 @@ message BlockOffset {
 message TableIndex {
   repeated BlockOffset offsets = 1;
   bytes bloom_filter = 2;
-  uint64 estimated_size = 3;
+  uint32 estimated_size = 3;
```
skyzh, line 58: Current badger master uses u32.

src/format.rs
```diff
@@ -9,16 +9,10 @@ pub fn key_with_ts(key: impl Into<BytesMut>, ts: u64) -> Bytes {
...
-            buf.as_mut_ptr() as *mut _,
-            8,
-        );
-        key.advance_mut(8);
-    }
+    let mut ts_bytes = key.split_off(key.len());
+    let res = u64::MAX - ts;
+    ts_bytes.put_u64(res);
+    key.unsplit(ts_bytes);
 }
```
skyzh, line 16: Here I rewrote `append_ts` and `get_ts` in a safe approach. Benchmark result shows that safe version is nearly the same performance as previous unsafe version, and safe `get_ts` is 2x slower than unsafe version. So I retained previous `get_ts` and put safe `append_ts` here.

image

src/table.rs
```diff
@@ -1,5 +1,5 @@
 mod builder;
-
+/*
```
skyzh, line 2: I'll fix this module later.

src/table/builder.rs
```diff
@@ -114,36 +136,45 @@ impl Builder {
...
+        let mut bytes = BytesMut::new();
+        self.table_index.encode(&mut bytes).unwrap();
         assert!(bytes.len() < u32::MAX as usize);
         self.buf.put_slice(&bytes);
-        self.buf.put_u32_le(bytes.len());
+        self.buf.put_u32_le(bytes.len() as u32);
         let cs = self.build_checksum(&bytes);
         self.write_checksum(cs);
-        self.buf.freeze()
+        self.buf.clone().freeze()
```
skyzh, line 162: This freeze may introduce overhead. I'll investigate how to fix this later.

src/format.rs
BusyJay: I think there are problems with criterion or the compiler. I can get an opposite result after moving the legacy case as the first case to bench.

src/format.rs
```diff
@@ -9,16 +9,10 @@ pub fn key_with_ts(key: impl Into<BytesMut>, ts: u64) -> Bytes {
...
-    let buf = key.bytes_mut();
-    unsafe {
-        ptr::copy_nonoverlapping(
-            &res as *const u64 as *const u8,
-            buf.as_mut_ptr() as *mut _,
-            8,
-        );
-        key.advance_mut(8);
-    }
+    let mut ts_bytes = key.split_off(key.len());
```
BusyJay, line 12: I don't think split is necessary.

src/table/builder.rs
skyzh: Freezing a `BytesMut` requires full ownership of `buf`. But in `finish(&mut self)`, we could only get a `&mut buf`. I'm not sure whether calling `finish` allows us to give away `builder`. I'll change this to `finish(self)` and remove `clone()` later, if `builder` won't be used after `finish`.

Cargo.toml
```diff
@@ -15,8 +15,15 @@ memmap = "0.7"
...
+[dev-dependencies]
+criterion = "0.3"
+
 [workspace]
 members = [
     "proto",
     "skiplist",
 ]
+
+[[bench]]
```
BusyJay, line 27: Should add a new profile for bench. `codegen-units` should be set to 1 to get a correct bench result.

BusyJay (commented):
Rest LGTM

Cargo.toml
skyzh: I copied TiKV `[profile.bench]` to `Cargo.toml`

BusyJay (approved):

PR #14, opened 2020-10-13
add table builder test

skyzh:
This PR adds (placeholders of) table builder tests. `TODO`s will be filled in after full table module is complete.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,10 +7,10 @@ pub mod format;
 mod levels;
 mod memtable;
 mod ops;
-mod opt;
-mod table;
+pub mod opt;
+pub mod table;
 mod util;
-mod value;
+pub mod value;
 mod wal;
 
 pub use db::{Agate, AgateOptions};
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -6,10 +6,10 @@ pub mod format;
 mod levels;
 mod memtable;
 mod ops;
-mod opt;
-mod table;
+pub mod opt;
```
BusyJay, line 9: Better reexport them instead of making internal module public.

src/table/builder.rs
```diff
@@ -177,3 +178,68 @@ impl Builder {
...
+        // TODO: use cache
+        let opts = Options {
+            block_size: 4 * 1024,
+            bloom_false_positive: 0.01,
+            table_size: 30 << 20,
+        };
+
+        let mut builder = Builder::new(opts);
+        let tmp_dir = TempDir::new("agatedb").unwrap();
+        let _filename = tmp_dir.path().join(rand_string());
```
BusyJay, line 209: You don't need to use `rand_string`, `TempDir` handles it already.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

PR #17, opened 2020-10-13
add sanitizer

skyzh:

fix https://github.com/tikv/agatedb/issues/10

Sanitizer will only run when there's any change in `skiplist` module. First successful run is triggered along with `split sanitizer` commit.

In the future, we could apply sanitizer to other modules.


Makefile
```diff
@@ -13,13 +10,16 @@ clippy:
...
+	RUSTFLAGS="$(SANITIZER_FLAGS)" cargo test --all-features --workspace
+
 bench:
 	cargo bench --all-features --workspace
 
-ci: fmt clippy test run_release
+bench_sanitizer:
+	RUSTFLAGS="$(SANITIZER_FLAGS)" cargo bench --all-features --workspace
 
-proto: proto/proto/meta.proto proto/proto/rustproto.proto
```
BusyJay, line 21: Why remove this?

BusyJay (commented):
Why limit to skiplist only?

Makefile
skyzh: Now we use prost-build to generate `meta.rs` on cargo build, instead of generating them manually.

skyzh:
> Why limit to skiplist only?

Run all sanitizer bench takes ~40mins, and unsafe code is mostly used in skiplists. So I choose to run sanitizer bench only on there's any change in skiplist.

Maybe we could run this bench manually instead of triggering by PR or commit, and I could include all modules in bench?

BusyJay:
> Maybe we could run this bench manually

Seems more practical.

skyzh:
I've updated the CI config. After this PR being merged, we could manually trigger sanitizer test and bench on GitHub Action webpage.

BusyJay (approved):

skyzh:
image

Now benchmark can be triggered in Action tab

PR #18, opened 2020-10-16
table: support checksum modes

skyzh:
Will be ready for review when table implementation has been merged.

I plan to support all badger checksum options `ChecksumVerificationMode` in this PR

```go
const (
	// NoVerification indicates DB should not verify checksum for SSTable blocks.
	NoVerification ChecksumVerificationMode = iota
	// OnTableRead indicates checksum should be verified while opening SSTtable.
	OnTableRead
	// OnBlockRead indicates checksum should be verified on every SSTable block read.
	OnBlockRead
	// OnTableAndBlockRead indicates checksum should be verified
	// on SSTable opening and on every block read.
	OnTableAndBlockRead
)
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,7 +7,7 @@ mod format;
 mod levels;
 mod memtable;
 mod ops;
-mod opt;
+pub mod opt;
 mod table;
 mod util;
 mod value;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -14,8 +14,8 @@ mod value;
 mod wal;
 
 pub use format::{get_ts, key_with_ts};
-pub use opt::Options as TableOptions;
 pub use opt::ChecksumVerificationMode;
+pub use opt::Options as TableOptions;
 pub use table::builder::Builder as TableBuilder;
 pub use table::Table;
 pub use value::Value;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,7 +7,7 @@ mod format;
 mod levels;
 mod memtable;
 mod ops;
-pub mod opt;
+mod opt;
 mod table;
 mod util;
 mod value;
@@ -15,6 +15,7 @@ mod wal;
 
 pub use format::{get_ts, key_with_ts};
 pub use opt::Options as TableOptions;
+pub use opt::ChecksumVerificationMode;
 pub use table::builder::Builder as TableBuilder;
 pub use table::Table;
 pub use value::Value;
--- a/src/table.rs
+++ b/src/table.rs
@@ -2,7 +2,7 @@ pub(crate) mod builder;
 mod iterator;
 
 use crate::checksum;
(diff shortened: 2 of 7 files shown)


skyzh:
@Connor1996 @BusyJay PTAL, thanks!

src/table.rs
```diff
@@ -256,16 +271,24 @@ impl TableInner {
...
-            // The checksum is calculated for actual data + entry index + index length
-            data: data.slice(..read_pos + 4),
+            data,
             entry_offsets,
             checksum_len,
             checksum,
-        }))
+        });
+
+        if matches!(self.opts.checksum_mode, OnTableAndBlockRead | OnTableRead) {
```
Connor1996, line 287:
```suggestion
        if matches!(self.opts.checksum_mode, OnTableAndBlockRead | OnBlockRead) {
```

Connor1996 (approved):
LGTM

src/table.rs
```diff
@@ -126,19 +128,25 @@ impl TableInner {
             index: TableIndex::default(),
             index_start: 0,
             index_len: 0,
-            opts,
+            opts: opts.clone(),
         };
         inner.init_biggest_and_smallest()?;
-        // TODO: verify checksum
+
+        if matches!(opts.checksum_mode, OnTableAndBlockRead | OnTableRead) {
```
BusyJay, line 135: Why not `inner.opts.checksum_mode`? So you don't need to clone it at L131.

src/lib.rs
```diff
@@ -7,7 +7,7 @@ mod format;
 mod levels;
 mod memtable;
 mod ops;
-mod opt;
+pub mod opt;
```
BusyJay, line 10: How about also reexporting it? So users don't need to write two lines for importing from agatedb.

src/table/tests.rs
```diff
@@ -344,3 +356,17 @@ fn test_table_big_values() {
     // TODO: support max_version in table
     // assert_eq!(n, table.max_version());
 }
+
+#[test]
+fn test_table_checksum() {
+    let mut rng = thread_rng();
+    let opts = get_test_table_options();
```
BusyJay, line 363: Original implement seems to use `OnTableAndBlockRead`.

skyzh:
@BusyJay PTAL, thanks!

src/table.rs
```diff
@@ -150,6 +158,11 @@ impl TableInner {
             index_len: 0,
         };
         inner.init_biggest_and_smallest()?;
+
+        if matches!(opts.checksum_mode, OnTableAndBlockRead | OnTableRead) {
```
BusyJay, line 162: You can still use `inner.opts` to get rid of `opts.clone()`.

BusyJay (commented):
Rest LGTM

BusyJay (approved):

PR #19, opened 2020-10-20
table: prepare for adding table impl

skyzh:

* fix typo of `Comparator`
* fix endian for all SST
* prepare for adding table implementation

### DIFF
--- a/skiplist/src/key.rs
+++ b/skiplist/src/key.rs
@@ -1,23 +1,23 @@
 use bytes::Bytes;
 use std::cmp::Ordering;
 
-pub trait KeyComparitor {
+pub trait KeyComparator {
     fn compare_key(&self, lhs: &[u8], rhs: &[u8]) -> Ordering;
     fn same_key(&self, lhs: &[u8], rhs: &[u8]) -> bool;
 }
 
 #[derive(Default, Debug, Clone, Copy)]
-pub struct FixedLengthSuffixComparitor {
+pub struct FixedLengthSuffixComparator {
     len: usize,
 }
 
-impl FixedLengthSuffixComparitor {
-    pub const fn new(len: usize) -> FixedLengthSuffixComparitor {
-        FixedLengthSuffixComparitor { len }
+impl FixedLengthSuffixComparator {
+    pub const fn new(len: usize) -> FixedLengthSuffixComparator {
+        FixedLengthSuffixComparator { len }
     }
 }
 
-impl KeyComparitor for FixedLengthSuffixComparitor {
+impl KeyComparator for FixedLengthSuffixComparator {
     #[inline]
     fn compare_key(&self, lhs: &[u8], rhs: &[u8]) -> Ordering {
         if lhs.len() < self.len {
--- a/skiplist/src/key.rs
+++ b/skiplist/src/key.rs
@@ -1,3 +1,4 @@
+use bytes::Bytes;
 use std::cmp::Ordering;
 
 pub trait KeyComparitor {
@@ -11,7 +12,7 @@ pub struct FixedLengthSuffixComparitor {
 }
 
 impl FixedLengthSuffixComparitor {
-    pub fn new(len: usize) -> FixedLengthSuffixComparitor {
+    pub const fn new(len: usize) -> FixedLengthSuffixComparitor {
         FixedLengthSuffixComparitor { len }
     }
 }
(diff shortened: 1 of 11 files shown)


src/util.rs
```diff
@@ -33,3 +35,21 @@ pub fn bytes_diff<'a, 'b>(base: &'a [u8], target: &'b [u8]) -> &'b [u8] {
         target.get_unchecked(end..)
     }
 }
+
+/// simple rewrite of golang sort.Search
+pub fn search<F>(n: usize, mut f: F) -> usize
```
Connor1996, line 40: can we use `slice.binary_search_by` directly?

skiplist/src/key.rs
```diff
@@ -11,14 +12,28 @@ pub struct FixedLengthSuffixComparitor {
 }
 
 impl FixedLengthSuffixComparitor {
-    pub fn new(len: usize) -> FixedLengthSuffixComparitor {
+    pub const fn new(len: usize) -> FixedLengthSuffixComparitor {
         FixedLengthSuffixComparitor { len }
     }
 }
 
 impl KeyComparitor for FixedLengthSuffixComparitor {
```
Connor1996, line 20: typo? Should `KeyComparitor`  be `KeyComparator`?

src/table/builder.rs
```diff
@@ -4,20 +4,22 @@ use crate::{checksum, util};
...
 use proto::meta::{checksum::Algorithm as ChecksumAlg, BlockOffset, Checksum, TableIndex};
-use std::{u16, u32};
 
 #[repr(C)]
-struct Header {
-    overlap: u16,
-    diff: u16,
+#[derive(Default)]
+pub struct Header {
+    pub overlap: u16,
```
Connor1996, line 11: Please add the comments

src/util.rs
skyzh: The goal of using this customized search function is that,

* We need to search on a number range `0..n`, instead of on a slice of concrete data. Data will be loaded into memory upon each search iteration.
* It seems that we could not directly use `binary_search_by` on a `Range` object.

skyzh: FYI, https://github.com/tikv/agatedb/blob/462e28dfb5445d1f2a97a7cfee02cd51c88dc87a/src/table/iterator.rs#L133

skiplist/src/key.rs
skyzh: It seems so, I'll fix.

Connor1996 (approved):
LGTM

src/table/builder.rs
```diff
@@ -91,9 +100,9 @@ impl Builder {
             return;
         }
         for offset in &self.entry_offsets {
-            self.buf.put_u32_le(*offset);
+            self.buf.put_u32(*offset);
```
BusyJay, line 103: Why change this?

src/util.rs
```diff
@@ -33,3 +35,21 @@ pub fn bytes_diff<'a, 'b>(base: &'a [u8], target: &'b [u8]) -> &'b [u8] {
...
+where
+    F: FnMut(usize) -> bool,
+{
+    let mut i = 0;
+    let mut j = n;
+    while i < j {
+        let h = (i + j) >> 1;
```
BusyJay, line 47: " / 2" is more readable.

src/util.rs
```diff
@@ -33,3 +35,21 @@ pub fn bytes_diff<'a, 'b>(base: &'a [u8], target: &'b [u8]) -> &'b [u8] {
...
+        if !f(h) {
```
BusyJay, line 48: Should there be a case for equal?

skyzh: This functions finds the first entry that matches predicate `f`, which only returns true or false, and is different from Rust's binary search. Generally, f will return true when current entry is equal to or greater than the target. It is possible to refactor this to use Rust's `Ordering::*`, but I'll do this later, when table implementation is complete.

src/table/builder.rs
skyzh: I'll change this back.

skyzh: And I'll use little endian for all data in SST, to have a unified interface.

skyzh: Now data in SST are all stored using little endian.

Connor1996: seems badger use big-endian? Though I don't know if there is any difference between big-endian and little-endian for encode/decode.

skyzh: Badger now uses flatbuffer, and it seems to be little endian now...

src/table/builder.rs
```diff
@@ -175,7 +195,7 @@ impl Builder {
         let len = res.len();
         assert!(len < u32::MAX as usize);
         self.buf.put_slice(&res);
-        self.buf.put_u32(len as u32);
+        self.buf.put_u32_le(len as u32);
```
BusyJay, line 198: I think this uses `BigEndian`.

BusyJay (commented):
Endianness changes the bytes order and may make result files not compatible with each other's implementations.

src/table/builder.rs
skyzh: Yes, you're right. I read badger `builder.go` again and find out that all builder encoding / decoding is using big endian. And for those using flatbuffers, they're little endian by default. I'll try making these compatible in future PRs.

skyzh: except that:
* `Header`, in badger it directly copy it to data, so it uses system endian.
* Vector data, for example, `offset`, it also uses system endian.

FYI, https://github.com/dgraph-io/badger/blob/master/y/y.go

BusyJay (approved):

PR #20, opened 2020-10-20
table: add bloom filter

skyzh:


src/bloom.rs
```diff
@@ -0,0 +1,104 @@
...
+    /// Get bloom filter bits per key from entries count and FPR
+    pub fn bloom_bits_per_key(entries: usize, false_positive_rate: f64) -> usize {
+        let size =
+            -1.0 * (entries as f64) * false_positive_rate.ln() / (0.69314718056 as f64).powi(2);
+        let locs = 0.69314718056 * size / (entries as f64);
+        locs as usize
+    }
+
+    /// Build bloom filter from key hashes
+    pub fn build_from_key_hashes(keys: &[u32], bits_per_key: usize) -> Bytes {
```
Connor1996, line 31: can we make it return `Self` directly?
and rename `new` into `from_bytes`?

skyzh: `Bloom` do not own any data by design. So we could not return `Self`.

skyzh: The bloom filter is stored inside `TableIndex`, which is a protobuf struct. Therefore, we we are accessing table index, we could only get a byte sequence representing a bloom filter. By using current scheme, we could construct a bloom without any copy.

src/table.rs
```diff
@@ -197,7 +202,8 @@ impl TableInner {
         // TODO: compression
         self.estimated_size = self.table_size as u32;
 
-        // TODO: has bloom filter
+        // bloom filter
+        self.has_bloom_filter = self.index.bloom_filter.len() > 0;
```
BusyJay, line 206: You mean `!self.index.bloom_filter.is_empty()`?

Cargo.toml
```diff
@@ -14,6 +14,7 @@ skiplist = { path = "skiplist" }
 memmap = "0.7"
 farmhash = "1.1"
 prost = "0.6"
+bitvec = "0.19"
```
BusyJay, line 17: I suggest not to use `bitvec` as the code is quite simple and `bitvec` can bring another 4 indirect deps.

skyzh: I've implemented a simple bitvec.

src/bloom.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// BitSlice allows getting individual bits on a u8 slice
+pub struct BitSlice<'a>(&'a [u8]);
+
+impl<'a> BitSlice<'a> {
+    pub fn new(data: &'a [u8]) -> Self {
+        Self(data)
+    }
+
+    pub fn len(&self) -> usize {
```
BusyJay, line 11: You can define them as trait and implement for `AsRef<[u8]>`.

src/bloom.rs
```diff
@@ -0,0 +1,155 @@
...
+impl BitVec {
+    pub fn with_capacity(nbits: usize) -> Self {
+        assert!(nbits % 8 == 0);
+        let vec_size = nbits / 8;
+        let mut data = Vec::with_capacity(vec_size);
+        data.resize(vec_size, 0);
+        Self(data)
+    }
+
+    pub fn set(&mut self, idx: usize, val: bool) {
```
BusyJay, line 34: This can be implement for `AsMut<[u8]>`.

src/bloom.rs
```diff
@@ -0,0 +1,133 @@
...
+    pub fn new(buf: &'a [u8]) -> Self {
+        let filter = &buf[..buf.len() - 4];
+        let k = (&buf[buf.len() - 4..]).get_u32();
+        Self { filter, k }
+    }
+
+    /// Get bloom filter bits per key from entries count and FPR
+    pub fn bloom_bits_per_key(entries: usize, false_positive_rate: f64) -> usize {
+        let size =
+            -1.0 * (entries as f64) * false_positive_rate.ln() / (0.69314718056 as f64).powi(2);
```
BusyJay, line 54: You can write `0.69314718056f64`

src/bloom.rs
```diff
@@ -0,0 +1,133 @@
...
+        let locs = 0.69314718056 * size / (entries as f64);
+        locs as usize
+    }
+
+    /// Build bloom filter from key hashes
+    pub fn build_from_key_hashes(keys: &[u32], bits_per_key: usize) -> Bytes {
+        let k = ((bits_per_key as f64) * 0.69) as u32;
+        let k = k.min(30).max(1);
+        let nbits = (keys.len() * bits_per_key).max(64);
+        let nbytes = (nbits + 7) >> 3;
```
BusyJay, line 64: Use `/ 8`, `* 8` instead for better reading.

src/bloom.rs
```diff
@@ -0,0 +1,133 @@
...
+        // nbits is always multiplication of 8
+        let nbits = nbytes << 3;
+        let mut filter = Vec::with_capacity(nbytes);
```
BusyJay, line 67: Why not use `BytesMut` in the first place?

src/bloom.rs
```diff
@@ -0,0 +1,133 @@
...
+            let delta = (h >> 17) | (h << 15);
+            for _ in 0..k {
+                let bit_pos = (h as usize) % nbits;
+                filter.set_bit(bit_pos, true);
+                h = h.wrapping_add(delta);
+            }
+        }
+        let mut buf = BytesMut::new();
+        buf.put_slice(filter.as_slice());
+        buf.put_u32(k);
```
BusyJay, line 80: It seems in `[1, 30]`.

skyzh: k is set to be in `[1, 30]` at the beginning of this func by `let k = k.min(30).max(1);`

Connor1996 (approved):
LGTM

skyzh:
@BusyJay PTAL, thanks!

BusyJay (commented):
Rest LGTM

src/bloom.rs
```diff
@@ -0,0 +1,134 @@
...
+        for h in keys {
+            let mut h = *h;
+            let delta = (h >> 17) | (h << 15);
+            for _ in 0..k {
+                let bit_pos = (h as usize) % nbits;
+                filter.set_bit(bit_pos, true);
+                h = h.wrapping_add(delta);
+            }
+        }
+        filter.put_u32(k);
```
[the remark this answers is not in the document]
skyzh: It seems that badger would use `k > 30` for new encodings. For now, I have changed that to use `u8`.

BusyJay (approved):

PR #21, opened 2020-10-20
table: add block iterator

skyzh:


src/table.rs
```diff
@@ -45,74 +65,14 @@ impl Drop for TableInner {
...
-        (0, false)
-    }
-    match name[..name.len() - 4].parse() {
-        Ok(id) => (id, true),
-        Err(_) => return (0, false),
-    }
-}
-
-impl Table {
-    pub fn open(path: &Path, opt: Options) -> Result<Table> {
```
BusyJay, line 82: Why delete it?

skyzh: I've added them back. In fact this Table implementation differs a lot from the final one, I think we could compare them in next few PRs.

src/table/iterator.rs
```diff
@@ -0,0 +1,170 @@
...
+    // TODO: As we need to clone Error from block iterator to table iterator,
+    // we had to save `crate::Error` as String. In the future, we could let all
+    // seek-related function to return a `Result<()>` instead of saving an
+    // error inside struct.
+    Error(String),
+}
+
+impl IteratorError {
+    pub fn is_err(&self) -> bool {
+        match self {
```
BusyJay, line 21: Perhaps you can use `matches!`.

src/table/iterator.rs
```diff
@@ -0,0 +1,170 @@
...
+enum SeekPos {
+    Origin,
+    Current,
+}
+
+/// Block iterator iterates on an SST block
+// TODO: support custom comparator
+struct BlockIterator {
+    idx: isize,
+    base_key: Bytes,
```
BusyJay, line 44: How about using `BytesMut` instead? So you can clone the diff key without allocating frequently.

src/table.rs
```diff
@@ -1,39 +1,59 @@
...
+impl MmapFile {
+    pub fn is_in_memory(&self) -> bool {
+        match self {
+            Self::File { .. } => false,
+            Self::Memory { .. } => true,
+        }
+    }
 }
 
 pub struct TableInner {
```
Connor1996, line 38: Add comments by the way

src/table/iterator.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    fn set_idx(&mut self, i: isize) {
+        self.idx = i;
+        if i >= self.entry_offsets().len() as isize || i < 0 {
+            self.err = IteratorError::EOF;
+            return;
+        }
+
+        self.err = IteratorError::NoError;
```
Connor1996, line 81: can we make err to be `Option<>`?

src/table/iterator.rs
```diff
@@ -0,0 +1,170 @@
...
+}
+
+enum SeekPos {
+    Origin,
+    Current,
+}
+
+/// Block iterator iterates on an SST block
+// TODO: support custom comparator
+struct BlockIterator {
```
Connor1996, line 42: Please add the comment of fields

src/table/iterator.rs
```diff
@@ -0,0 +1,170 @@
...
+        self.idx = i;
+        if i >= self.entry_offsets().len() as isize || i < 0 {
+            self.err = IteratorError::EOF;
+            return;
+        }
+
+        self.err = IteratorError::NoError;
+
+        // if base key is not available, retrieve it from first key-value pair in block
+        if self.base_key.is_empty() {
```
Connor1996, line 77: use `Option<>`?

src/table/iterator.rs
skyzh: I've changed `key` to `BytesMut`, and now we could construct key without allocating new space.

src/table/iterator.rs
skyzh: This was intended, as originally I thought these two structs will take different spaces. Just now I found that `Option<Iterator>` takes the same space as current implementation. I'll change to `Option<>`

image

src/table/iterator.rs
```diff
@@ -0,0 +1,179 @@
...
+        // TODO: merge this truncate with the following key truncate
+        if header.overlap > self.perv_overlap {
+            self.key.truncate(self.perv_overlap as usize);
+            self.key.extend_from_slice(
+                &self.base_key[self.perv_overlap as usize..header.overlap as usize],
+            );
+        }
+        self.perv_overlap = header.overlap;
+
+        let diff_key = &entry_data[..header.diff as usize];
```
Connor1996, line 105: where is `valueOff`?

src/table/iterator.rs
skyzh: FYI, now it takes 32 bytes.

src/table/iterator.rs
skyzh: As `header.decode(&mut entry_data);` eats headers from `entry_data`, at this point, `entry_data` won't contain header. So we don't need to introduce `valueOff`.

src/table/iterator.rs
```diff
@@ -0,0 +1,180 @@
...
+        self.key.truncate(header.overlap as usize);
+        self.key.extend_from_slice(diff_key);
+        self.val = entry_data.slice(header.diff as usize..);
+    }
+    /// Check if last operation of iterator is error
```
Connor1996, line 117: Add a blank

Connor1996 (approved):
rest LGTM

src/table/builder.rs
```diff
@@ -22,7 +22,7 @@ impl Header {
         bytes.put_u32_le((self.overlap as u32) << 16 | self.diff as u32);
     }
 
-    pub fn decode(&mut self, bytes: &mut Bytes) {
+    pub fn decode(&mut self, mut bytes: &[u8]) {
```
BusyJay, line 25: Why change?

src/table/iterator.rs
```diff
@@ -0,0 +1,181 @@
+use super::builder::{Header, HEADER_SIZE};
+use super::{Block, TableInner};
```
BusyJay, line 2: Please fix the warnings.

src/table/builder.rs
skyzh: It seems that we could retain original design once we changed the way of constructing key in previous commits. I've changed that back.

src/table/iterator.rs
```diff
@@ -0,0 +1,180 @@
...
+    block: Arc<Block>,
+    /// previous overlap key, used to construct key of current entry from
+    /// previous one faster
+    perv_overlap: u16,
+    /// iterator error in last operation
+    err: Option<IteratorError>,
+}
+
+impl BlockIterator {
+    pub fn new(block: Arc<Block>) -> Self {
```
BusyJay, line 54: Provide `set_block` like upstream can also reduce allocations.

src/table/iterator.rs
```diff
@@ -0,0 +1,180 @@
...
+            idx: 0,
+        }
+    }
+
+    #[inline]
+    fn entry_offsets(&self) -> &[u32] {
+        &self.block.entry_offsets
+    }
+
+    fn set_idx(&mut self, i: isize) {
```
BusyJay, line 73: Can it use usize?

BusyJay (commented):
Rest LGTM

src/table/iterator.rs
skyzh: badger will invalidate the iterator when index is negative. We could use `usize` in parameter, but will have to store `isize` inside the iterator.

skyzh: Hmm... I think it's better to use `isize` here, as `self.idx` should always contain the same value as input parameter. When user provides a `usize` larger than `isize` max, there could be inconsistency. So I would prefer using `isize` in all iterators.

BusyJay: Why not make `self.idx` as `usize`?

skyzh: I've refactored iterator to store `usize`, and added extra check for reaching reverse iterator end in `seek_last`.

src/table/iterator.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    fn set_idx(&mut self, i: usize) {
+        self.idx = i;
+        if i >= self.entry_offsets().len() {
+            self.err = Some(IteratorError::EOF);
+            return;
+        }
+
+        self.err = None;
+        let start_offset = self.entry_offsets()[i as usize] as u32;
```
BusyJay, line 93: Cast is unnecessary anymore.

BusyJay (approved):

PR #22, opened 2020-10-20
table: implement table iterator

skyzh:

Based on https://github.com/tikv/agatedb/pull/21

### DIFF
--- a/src/table/iterator.rs
+++ b/src/table/iterator.rs
@@ -233,6 +233,8 @@ impl<T: AsRef<TableInner>> Iterator<T> {
         self.err = None;
     }
 
+    /// Check if last operation of iterator is error
+    /// TODO: use `Result<()>` for all iterator operation and remove this if possible
     pub fn valid(&self) -> bool {
         self.err.is_none()
     }
@@ -387,6 +389,8 @@ impl<T: AsRef<TableInner>> Iterator<T> {
             if !bi.valid() {
                 if self.bpos == 0 {
                     self.bpos = std::usize::MAX;
+                    // At this point, the iterator must be reset before using.
+                    // If the user calls `next`, `bpos` will overflow and panic.
                     self.err = Some(IteratorError::EOF);
                     return;
                 }
--- a/src/table/iterator.rs
+++ b/src/table/iterator.rs
@@ -376,7 +376,7 @@ impl<T: AsRef<TableInner>> Iterator<T> {
             bi.next();
             if !bi.valid() {
                 self.bpos += 1;
-                bi.data = Bytes::new();
+                bi.data.clear();
                 self.next_inner();
             }
         }
@@ -408,7 +408,7 @@ impl<T: AsRef<TableInner>> Iterator<T> {
             bi.prev();
             if !bi.valid() {
                 self.bpos -= 1;
-                bi.data = Bytes::new();
+                bi.data.clear();
                 self.prev_inner();
             }
         }
--- a/src/table/iterator.rs
+++ b/src/table/iterator.rs
@@ -200,6 +200,13 @@ impl BlockIterator {
             self.set_idx(self.idx - 1);
         }
     }
+
+    pub fn require_load(iter: &Option<Self>) -> bool {
+        match iter {
+            Some(iter) => iter.data.is_empty(),
+            None => true,
+        }
+    }
 }
 
 // TODO: use `bitfield` if there are too many variants
@@ -355,7 +362,7 @@ impl<T: AsRef<TableInner>> Iterator<T> {
             return;
         }
 
-        if self.block_iterator.is_none() {
+        if BlockIterator::require_load(&self.block_iterator) {
             match self.table.as_ref().block(self.bpos, self.use_cache()) {
                 Ok(block) => {
                     let block_iterator = self.get_block_iterator(block);
@@ -378,7 +385,7 @@ impl<T: AsRef<TableInner>> Iterator<T> {
     fn prev_inner(&mut self) {
         self.err = None;
 
-        if self.block_iterator.is_none() {
+        if BlockIterator::require_load(&self.block_iterator) {
             match self.table.as_ref().block(self.bpos, self.use_cache()) {
                 Ok(block) => {
                     let block_iterator = self.get_block_iterator(block);
--- a/src/table/iterator.rs
+++ b/src/table/iterator.rs
@@ -241,8 +241,12 @@ impl<T: AsRef<TableInner>> Iterator<T> {
     fn get_block_iterator(&mut self, block: Arc<Block>) -> &mut BlockIterator {
         if self.block_iterator.is_none() {
             self.block_iterator = Some(BlockIterator::new(block));
+            self.block_iterator.as_mut().unwrap()
+        } else {
+            let iter = self.block_iterator.as_mut().unwrap();
+            iter.set_block(block);
+            iter
         }
-        self.block_iterator.as_mut().unwrap()
     }
 
     pub fn seek_to_first(&mut self) {
--- a/src/table/iterator.rs
+++ b/src/table/iterator.rs
@@ -1,6 +1,7 @@
 use super::builder::{Header, HEADER_SIZE};
-use super::Block;
+use super::{Block, TableInner};
 use crate::util::{self, KeyComparator, COMPARATOR};
+use crate::value::Value;
 use bytes::{Bytes, BytesMut};
 use std::sync::Arc;
 
--- a/src/table/iterator.rs
+++ b/src/table/iterator.rs
@@ -238,6 +238,13 @@ impl<T: AsRef<TableInner>> Iterator<T> {
         self.opt & ITERATOR_NOCACHE == 0
     }
 
+    fn get_block_iterator(&mut self, block: Arc<Block>) -> &mut BlockIterator {
+        if self.block_iterator.is_none() {
+            self.block_iterator = Some(BlockIterator::new(block));
+        }
+        self.block_iterator.as_mut().unwrap()
+    }
+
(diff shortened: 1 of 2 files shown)


src/table/iterator.rs
```diff
@@ -189,6 +195,269 @@ impl BlockIterator {
...
+        } else {
+            self.prev_inner();
+        }
+    }
+
+    /// `prev` points the iterator to previous element.
+    /// Note that if the iterator becomes invalid after operation,
+    /// you must reset the iterator by using `rewind` or `seek`
+    /// before using it again.
+    pub(crate) fn prev(&mut self) {
```
Connor1996, line 430: seems upstream doesn't provide `prev` in the iterator interface.

src/table/iterator.rs
```diff
@@ -189,6 +195,269 @@ impl BlockIterator {
...
+        }
+
+        if self.block_iterator.as_ref().unwrap().data.is_empty() {
+            match self.table.as_ref().block(self.bpos, self.use_cache()) {
+                Ok(block) => {
+                    let block_iterator = self.get_block_iterator(block);
+                    block_iterator.seek_to_first();
+                    self.err = block_iterator.err.clone();
+                }
+                Err(err) => self.err = Some(IteratorError::Error(err.to_string())),
```
Connor1996, line 361: how about using `?`, and impl `From<Error>` of the IteratorError?

src/table/iterator.rs
```diff
@@ -189,6 +195,269 @@ impl BlockIterator {
...
+
+    fn next_inner(&mut self) {
+        self.err = None;
+
+        if self.bpos > self.table.as_ref().offsets_length() {
+            self.err = Some(IteratorError::EOF);
+            return;
+        }
+
+        if self.block_iterator.as_ref().unwrap().data.is_empty() {
```
Connor1996, line 354: seems it should be `self.block_iterator.is_none()`

src/table/iterator.rs
```diff
@@ -189,6 +195,269 @@ impl BlockIterator {
...
+                bi.data.clear();
+                self.next_inner();
+            }
+        }
+    }
+
+    fn prev_inner(&mut self) {
+        self.err = None;
+
+        if self.block_iterator.as_ref().unwrap().data.is_empty() {
```
Connor1996, line 377: ditto

src/table/iterator.rs
skyzh: Table iterator provides a private `prev` for test use. FYI, https://github.com/dgraph-io/badger/blob/master/table/iterator.go#L331

src/table/iterator.rs
skyzh: Thanks for reminding. During my testing process, this branch never panics. I'll add test cases for direct calling `prev` / `next` without initializing iterator in next PR.

src/table/iterator.rs
skyzh: Seems that this function doesn't return `Result`, so I just called `Some(err.into())`

src/table/iterator.rs
```diff
@@ -189,6 +202,267 @@ impl BlockIterator {
...
+                block_iterator.seek(key, SeekPos::Origin);
+                self.err = block_iterator.err.clone();
+            }
+            Err(err) => self.err = Some(err.into()),
+        }
+    }
+
+    fn seek_from(&mut self, key: &Bytes, whence: SeekPos) {
+        self.err = None;
+        match whence {
```
BusyJay, line 310: Why not use `if`?

src/table/iterator.rs
```diff
@@ -189,6 +202,267 @@ impl BlockIterator {
...
+        Self {
+            table,
+            bpos: 0,
+            block_iterator: None,
+            err: None,
+            opt,
+        }
+    }
+
+    pub fn reset(&mut self) {
```
BusyJay, line 238: Seems should be private.

src/table/iterator.rs
```diff
@@ -189,6 +202,267 @@ impl BlockIterator {
...
+        match whence {
+            SeekPos::Origin => self.reset(),
+            _ => {}
+        }
+
+        let idx = util::search(self.table.as_ref().offsets_length(), |idx| {
+            use std::cmp::Ordering::*;
+            let block_offset = self.table.as_ref().offsets(idx).unwrap();
+            match COMPARATOR.compare_key(&block_offset.key, &key) {
+                Less => false,
```
BusyJay, line 319: The original implement returns false for Equal too.

src/table/iterator.rs
```diff
@@ -189,6 +202,267 @@ impl BlockIterator {
...
+
+    fn next_inner(&mut self) {
+        self.err = None;
+
+        if self.bpos > self.table.as_ref().offsets_length() {
+            self.err = Some(IteratorError::EOF);
+            return;
+        }
+
+        if self.block_iterator.is_none() {
```
BusyJay, line 359: Should also check if data is empty.

skyzh: In this implementation, `block_iterator` is always initialized with data. I guess it is not necessary.

BusyJay: Then it should stack over flow when executing L373.

skyzh: Got it. I'll fix this soon.

skyzh: What about always giving the iterator an initial value? (e.g. points to the first block), then we won't need to deal with those Option<Iterator>s

BusyJay: I think there always be an `Option`, the different is whether the `Option` is pushed down. Though I'm OK with your proposal.

skyzh: Now we will load the block either the iterator isn't initialized, or the data is empty.

src/table/iterator.rs
Connor1996: it's `prev_inner`

Connor1996 (commented):
most LGTM

skyzh:
Two review comments have been solved,
* Now we will load the block either the iterator isn't initialized, or the data is empty.
* Remove `prev` from Iterator struct

Connor1996 (approved):

skyzh:
PTAL, thanks! @BusyJay

src/table/iterator.rs
```diff
@@ -187,6 +200,261 @@ impl BlockIterator {
...
+        self.seek_from(key, SeekPos::Origin);
+        if self.key() != key {
+            self.prev_inner();
+        }
+    }
+
+    fn next_inner(&mut self) {
+        self.err = None;
+
+        if self.bpos > self.table.as_ref().offsets_length() {
```
BusyJay, line 360: Should be `>=`.

src/table/iterator.rs
```diff
@@ -187,6 +200,261 @@ impl BlockIterator {
...
+            self.err = Some(IteratorError::EOF);
+            return;
+        }
+
+        if BlockIterator::require_load(&self.block_iterator) {
```
BusyJay, line 365: Should name it `is_xxx`.

src/table/iterator.rs
```diff
@@ -187,6 +200,261 @@ impl BlockIterator {
...
+                self.bpos += 1;
+                bi.data.clear();
+                self.next_inner();
+            }
+        }
+    }
+
+    fn prev_inner(&mut self) {
+        self.err = None;
+
```
BusyJay, line 387: Should check if `bpos == MAX`.

src/table/iterator.rs
```diff
@@ -187,6 +200,261 @@ impl BlockIterator {
...
+                Err(err) => self.err = Some(err.into()),
+            }
+        } else {
+            let bi = self.block_iterator.as_mut().unwrap();
+            bi.prev();
+            if !bi.valid() {
+                if self.bpos == 0 {
+                    self.bpos = std::usize::MAX;
+                    // At this point, the iterator must be reset before using.
+                    // If the user calls `next`, `bpos` will overflow and panic.
```
BusyJay, line 404: Should clear data too to reclaim memory early.

src/table/iterator.rs
```diff
@@ -187,6 +200,261 @@ impl BlockIterator {
...
+                    block_iterator.seek_to_last();
+                    self.err = block_iterator.err.clone();
+                }
+                Err(err) => self.err = Some(err.into()),
+            }
+        } else {
+            let bi = self.block_iterator.as_mut().unwrap();
+            bi.prev();
+            if !bi.valid() {
+                if self.bpos == 0 {
```
BusyJay, line 401: You can use `overflowing_sub` to avoid condition check.

src/table/iterator.rs
```diff
@@ -187,6 +200,261 @@ impl BlockIterator {
...
+                    return;
+                }
+                self.bpos -= 1;
+                bi.data.clear();
+                self.prev_inner();
+            }
+        }
+    }
+
+    pub fn key(&self) -> &[u8] {
```
BusyJay, line 415: How about returning `&Bytes`?

BusyJay (commented):
Rest LGTM

src/table/iterator.rs
skyzh: As `key` is stores as `BytesMut`, returning `Bytes` requires a clone.

src/table/iterator.rs
skyzh: Now I use `warpping_sub` and check if bpos is `MAX` upon entering `prev_inner`

src/table/iterator.rs
skyzh: So I think it's better to decide whether to clone when using this key, instead of always cloning it when getting the key.

BusyJay (approved):

PR #23, opened 2020-10-20
table: use mmap when reading

skyzh:

Will be ready for review when table implementation has been merged.

### DIFF
--- a/src/table.rs
+++ b/src/table.rs
@@ -7,10 +7,11 @@ use crate::Error;
 use crate::Result;
 use bytes::{Buf, Bytes};
 use iterator::{Iterator as TableIterator, ITERATOR_NOCACHE, ITERATOR_REVERSED};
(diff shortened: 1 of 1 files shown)


Connor1996 (approved):
LGTM

src/table.rs
```diff
@@ -24,8 +24,8 @@ mod tests;
 enum MmapFile {
     File {
         name: PathBuf,
-        // TODO: remove this mutex and allow multi-thread read
-        file: Mutex<fs::File>,
+        file: fs::File,
```
BusyJay, line 27: Why still keeping it?

skyzh: `Mmap` only needs reference to a file (as the OS just needs the file handler to create a mmap file). Therefore, we have to own the File struct inside `MmapFile`.

BusyJay (approved):

PR #24, opened 2020-10-21
skiplist: Fix undefined behavior in arena drop

brson:
The Vec must be deallocated with the same pointer alignment and capacity as it
is created.

Signed-off-by: Brian Anderson <andersrb@gmail.com>

Found with miri cc @RalfJung


BusyJay (approved):

BusyJay:
Perhaps we can also add miri check in CI.

skyzh (approved):
Thanks! I'll add miri check later

PR #25, opened 2020-10-21
table: implement table

skyzh:

Based on https://github.com/tikv/agatedb/pull/22

Tests will be added along with benchmarks in next PR.


src/table.rs
```diff
@@ -65,31 +70,146 @@ pub struct Table {
...
+    fn as_ref(&self) -> &TableInner {
+        self
+    }
+}
+
 impl TableInner {
     /// Create an SST from bytes data generated with table builder
-    pub fn create(_path: &Path, _data: Bytes, _opt: Options) -> Result<TableInner> {
-        unimplemented!()
+    fn create(path: &Path, data: Bytes, opt: Options) -> Result<TableInner> {
```
Connor1996, line 83:
```suggestion
    fn create(path: &Path, data: Bytes, opts: Options) -> Result<TableInner> {
```
same for other signatures

src/table.rs
```diff
@@ -65,31 +70,146 @@ pub struct Table {
...
+        let mut f = fs::OpenOptions::new()
+            .create_new(true)
+            .read(true)
+            .write(true)
+            .open(path)?;
+        f.write(&data)?;
+        // TODO: pass file object directly to open
+        drop(f);
```
Connor1996, line 91: // TODO: sync write

src/table.rs
```diff
@@ -65,31 +70,146 @@ pub struct Table {
...
+        // TODO: compression
+        self.estimated_size = self.table_size as u32;
+
+        // TODO: has bloom filter
+
+        Ok(&self.index.offsets[0])
     }
 
     fn key_splits(&mut self, _n: usize, _prefix: Bytes) -> Vec<String> {
+        // if n == 0 {
```
Connor1996, line 202: why comment out?

src/table.rs
```diff
@@ -106,8 +226,51 @@ impl TableInner {
...
+        let mut entry_offsets_ptr = &data[entries_index_start..entries_index_end];
+        let mut entry_offsets = Vec::with_capacity(num_entries);
+        for _ in 0..num_entries {
+            entry_offsets.push(entry_offsets_ptr.get_u32_le());
+        }
+
+        Ok(Arc::new(Block {
+            offset,
+            entries_index_start,
+            data: data.slice(..read_pos + 4),
```
Connor1996, line 269: Add comment 
```
// Drop checksum and checksum length.
// The checksum is calculated for actual data + entry index + index length
```

src/table.rs
skyzh: This function is not implemented, and will be implemented later.

src/table.rs
skyzh: Alright, I'll try to make them consistent

src/table.rs
skyzh: Here we dropped `f`, so I guess table writing should have been synchronized when dropping (in current situation). Whatever, I still added this todo.

src/table.rs
skyzh: I've removed the implementation for now

Connor1996 (approved):
LGTM

skyzh:
PTAL, thanks! @BusyJay

src/table.rs
BusyJay: > A successful close does not guarantee that the data has been successfully saved to disk, as the kernel defers writes. It is not common for a file system to flush the buffers when the stream is closed. If you need to be sure that the data is physically stored use fsync(2). (It will depend on the disk hardware at this point.)

No, file will not be synced when dropped. We have met the issue several times that data got lost without syncing them explicitly.

src/table.rs
```diff
@@ -169,19 +323,50 @@ impl TableInner {
...
+                    Ok(data.slice(offset..offset + size))
+                }
+            }
+            MmapFile::File { file, .. } => {
+                // TODO: use MmapFile
+                let mut file = file.lock().unwrap();
+                file.seek(std::io::SeekFrom::Start(offset as u64))?;
+                let mut buf = vec![0; size];
+                file.read_exact(&mut buf)?;
+                assert_eq!(buf.len(), size);
```
BusyJay, line 366: Redundant assert.

src/table.rs
```diff
@@ -58,35 +63,139 @@ pub struct TableInner {
...
 
     fn init_index(&mut self) -> Result<&BlockOffset> {
-        unimplemented!()
+        let mut read_pos = self.table_size;
+
+        // read checksum length from last 4 bytes
+        read_pos -= 4;
+        let mut buf = self.read(read_pos, 4)?;
+        let checksum_len = buf.get_u32() as usize;
+        if (checksum_len as i32) < 0 {
```
BusyJay, line 169: Is it necessary?

src/table.rs
```diff
@@ -58,35 +63,139 @@ pub struct TableInner {
...
+            index_len: 0,
+        };
+        inner.init_biggest_and_smallest()?;
+        Ok(inner)
     }
 
     fn init_biggest_and_smallest(&mut self) -> Result<()> {
-        unimplemented!()
+        let ko = self.init_index()?;
+        self.smallest = Bytes::copy_from_slice(&ko.key);
```
BusyJay, line 149: Why not just `clone`?

src/table.rs
```diff
@@ -218,64 +404,74 @@ pub struct Block {
...
+        3 * std::mem::size_of::<usize>() as u64
+            + self.data.len() as u64
+            + self.checksum.len() as u64
+            + self.entry_offsets.len() as u64 * std::mem::size_of::<u32>() as u64
     }
 
     fn verify_checksum(&self) -> Result<()> {
-        let mut chksum = CheckSum::default();
-        chksum.merge_from_bytes(&self.checksum)?;
+        let chksum = prost::Message::decode(self.data.clone())?;
```
BusyJay, line 416: I think it should be `self.checksum`.

src/table.rs
skyzh: Seems not necessary as we're using usize now.

src/table.rs
skyzh: Yes, I forget to cherry-pick https://github.com/tikv/agatedb/pull/18/ to this branch...

src/table.rs
```diff
@@ -58,35 +63,134 @@ pub struct TableInner {
...
+            index_len: 0,
+        };
+        inner.init_biggest_and_smallest()?;
+        Ok(inner)
     }
 
     fn init_biggest_and_smallest(&mut self) -> Result<()> {
-        unimplemented!()
+        let ko = self.init_index()?;
+        self.smallest = Bytes::from(ko.key.clone());
```
BusyJay, line 149: I thought it uses `Bytes` already, turned out I was wrong.

BusyJay (approved):

src/table.rs
skyzh: Yes, we have modified this in previous PR

PR #26, opened 2020-10-21
table: add tests and benchmarks

skyzh:
There are also some extra tests to add, and will be included in next PR, e.g.
* ensure that when iterator comes to end, its error is EOF instead of others
* ensure that the iterator functions correctly if it has not been reset
* ensure that the iterator can rewind if it exceeds the upper bound


src/table.rs
```diff
@@ -469,3 +469,353 @@ impl Table {
         self.inner.max_version()
     }
 }
+
+#[cfg(test)]
```
BusyJay, line 473: How about using separate file so that the current one won't get too long? You can do that by creating file src/table/tests.rs.

skyzh: Alright. But we may have to export some functions as `pub(crate)` in this case.

BusyJay: I don't think so. Access level is based on model instead of files.

skyzh: I've moved that to a separate file.

benches/bench.rs
```diff
@@ -1,7 +1,7 @@
-use agatedb::{get_ts, key_with_ts, TableBuilder, TableOptions, Value};
-use bytes::Bytes;
+mod common;
```
Connor1996, line 1: rename filename to `bench_common.rs`?

Connor1996 (commented):
rest LGTM

src/table/tests.rs
```diff
@@ -0,0 +1,346 @@
...
+
+        let mut count = 0;
+        while it.valid() {
+            let v = it.value();
+            assert_eq!(count.to_string(), v.value);
+            assert_eq!(b'A', v.meta);
+            it.next();
+            count += 1;
+        }
+    }
```
Connor1996, line 187:
```suggestion
    }
    assert_eq!(n, count);
```

Connor1996 (commented):
rest LGTM

benches/bench_table.rs
```diff
@@ -0,0 +1,124 @@
...
+            for j in 0..KEY_COUNT {
+                builder.add(&key_list[j], vs.clone(), 0);
+            }
+            builder.finish()
+        });
+    });
+}
+
+fn get_table_for_benchmark(count: usize) -> Table {
+    let tmp_dir = TempDir::new("agatedb").unwrap();
```
BusyJay, line 40: This can leave dirty file in temporary directory on Windows.

src/table/tests.rs
```diff
@@ -0,0 +1,348 @@
...
+    it.seek_to_first();
+    assert!(it.valid());
+    assert_eq!(user_key(it.key()), &key(b"key", 0)[..]);
+}
+
+#[test]
+fn test_uni_iterator() {
+    let opts = get_test_table_options();
+    let table = build_test_table(b"key", 10000, opts);
+    {
```
BusyJay, line 276: We use space instead of `{` to separate block of code.

benches/bench_table.rs
skyzh: What about passing table in a closure, like this changes that will land in following PRs?

https://github.com/tikv/agatedb/blob/070a61113d82d2e111a86f2a99be11b8bc30691c/src/db.rs#L320

skyzh: As this issue also exists in tests, and using schemes like `with_table_test` requires large refactor, I would like to fix this in a separate PR.

BusyJay (approved):

skyzh:
PTAL, thanks! @Connor1996

Connor1996 (approved):

PR #27, opened 2020-10-21
ci: add miri

skyzh:


skyzh:
Seems that Miri doesn't work with file systems. I'll investigate this later.

PR #29, opened 2020-10-27
agate, util: fix varint decode and add header

skyzh:
The implementation of varint decode seems to be wrong, and I've rewritten it from the Golang version. 

Encode / decode has been moved to `util.rs`.

By the way, I also implemented decode / encode of WAL header.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -1,6 +1,8 @@
 pub use skiplist::{FixedLengthSuffixComparator, KeyComparator};
 use std::{cmp, ptr};
 pub static COMPARATOR: FixedLengthSuffixComparator = FixedLengthSuffixComparator::new(8);
+use bytes::BytesMut;
+use std::mem::MaybeUninit;
 
 unsafe fn u64(ptr: *const u8) -> u64 {
     ptr::read_unaligned(ptr as *const u64)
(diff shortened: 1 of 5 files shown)


src/util.rs
```diff
@@ -53,3 +55,66 @@ where
...
+        s += 7;
+        i += 1;
+        p = p.add(1);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_varint_decode_encode() {}
```
skyzh, line 119: I'll fix this soon

src/util.rs
```diff
@@ -53,3 +55,92 @@ where
...
+        (*(&mut *p).as_mut_ptr()) = (u as u8 & 0x7f) | 0x80;
+        p = p.add(1);
+        u >>= 7;
+        c += 1;
+    }
+    (*(&mut *p).as_mut_ptr()) = u as u8;
+    c + 1
+}
+
+pub unsafe fn decode_varint_uncheck(bytes: &[u8]) -> (u64, usize) {
```
BusyJay, line 96: You can write it in safe rust by using for loop.

https://github.com/BusyJay/pecan/blob/d4a17231253cd6b304a15d9d0f65edd5a0bcceab/utils/src/codec.rs#L54-L84 is an example.

src/wal.rs
```diff
@@ -1,10 +1,60 @@
...
+            + varint_len(self.expires_at as usize)
+            + varint_len(self.key_len as usize)
+            + varint_len(self.value_len as usize)
+    }
+
+    pub fn encode(&self, bytes: &mut BytesMut) {
+        let encoded_len = self.encoded_len();
+        bytes.reserve(encoded_len);
+        let read = unsafe {
+            let buf = bytes.bytes_mut();
```
BusyJay, line 28: Perhaps adding an if check before entering `get_unchecked_mut` is better.

src/wal.rs
```diff
@@ -1,10 +1,60 @@
...
+            *(*buf.get_unchecked_mut(0)).as_mut_ptr() = self.meta;
+            *(*buf.get_unchecked_mut(1)).as_mut_ptr() = self.user_meta;
+            let mut read = 2;
+            read += encode_varint_uncheck(buf.get_unchecked_mut(read..), self.key_len as u64);
+            read += encode_varint_uncheck(buf.get_unchecked_mut(read..), self.value_len as u64);
+            read += encode_varint_uncheck(buf.get_unchecked_mut(read..), self.expires_at);
+            read
+        };
+        assert_eq!(read, encoded_len);
```
BusyJay, line 37: `debug_assert_eq` should be enough.

src/util.rs
skyzh: I copied `codec.rs` and modified input parameter as `BytesMut`, plus adapting error type to agatedb.

src/wal.rs
skyzh: I've added the check `assert!(buf.len() >= encoded_len);`. Seems that `reserve` will always ensure the capacity is enough, I'm not sure whether this length check is necessary or not.

src/wal.rs
```diff
@@ -1,10 +1,70 @@
...
-    value_len: usize,
+    key_len: u32,
+    value_len: u32,
+    expires_at: u64,
+    meta: u8,
+    user_meta: u8,
+}
+
+impl Header {
+    pub fn encoded_len(&self) -> usize {
```
Connor1996, line 20: Please add the comments of the format

src/wal.rs
```diff
@@ -1,10 +1,70 @@
 use super::Result;
+use crate::util::binary::{
```
Connor1996, line 2: How about using a crate like `https://docs.rs/integer-encoding/2.1.1/integer_encoding/` instead of porting a go `binary`. /cc @BusyJay

BusyJay: The latest patch seems to copy the implementation from a library written by me. Only three functions are necessary and they are quite simple. I suggest to avoid adding new dependencies and delete unused functions in the futures.

skyzh: I'll remove zig-zag functions and signed integer functions from `binary.rs`

skyzh: Done. Now we only retain functions for varint u32 and u64.

skyzh:
@Connor1996 Now that `binary.rs` is less than 300 lines. I think it's okay to embed it in this project instead of importing new crates.

BusyJay (approved):

Connor1996 (approved):

PR #30, opened 2020-10-28
table: add merge iterator

skyzh:
Will be ready for review when tests are ready

<del>The code is quite hard to read and understand when using `enum` to represent all iterators, which can be found in https://github.com/tikv/agatedb/pull/30/commits/63dc49189f37102487b04630e066be8da6a8362f. So I decided to just use `Box<dyn Iterator>` and refactor it later.</del>


longfangsong:
You can also try https://gitlab.com/antonok/enum_dispatch

skyzh:
Thanks @longfangsong , now I've implemented merge iterator with enum_dispatch. It's really easy to use 👍

src/lib.rs
```diff
@@ -8,6 +8,7 @@ mod levels;
 mod memtable;
 mod ops;
 mod opt;
+mod structs;
```
BusyJay, line 11: What does `structs` mean?

src/table/merge_iterator.rs
```diff
@@ -0,0 +1,362 @@
...
+use crate::Value;
+
+type TableIterator = Iterator<Arc<TableInner>>;
+
+/// `Iterators` includes all iterator types for AgateDB.
+/// By packing them into an enum, we could reduce the
+/// overhead of dynamic dispatch.
+#[enum_dispatch(AgateIterator)]
+pub enum Iterators {
+    MergeIterator(MergeIterator),
```
BusyJay, line 18: You can mark it `Box` to get rid of the `Box` at L40.

src/lib.rs
skyzh: structure-s? What about moving `AgateIterator` in `structs` to a separate mod?

src/table/merge_iterator.rs
skyzh: I tried using something like this:

```
#[enum_dispatch(Box<AgateIterator>)]
pub enum Iterators {
    MergeIterator(Box<MergeIterator>),
    TableIterator(Box<TableIterator>),
    #[cfg(test)]
    VecIterator(Box<tests::VecIterator>),
}
```

But this breaks method delegation (e.g. `iter.key()` won't work) and breaks `Iterators::from(TableIterator)`. Let me know if I misunderstood something.

src/lib.rs
BusyJay: Perhaps give it a more meaningful name. Memtable, Skiplist are structs too.

skyzh: I have renamed that to `iterator_trait`.

src/iterator_trait.rs
```diff
@@ -0,0 +1,15 @@
+use crate::value::Value;
+use bytes::Bytes;
+use enum_dispatch::enum_dispatch;
+
+/// `AgateIterator` defines the interface of all iterators,
+/// including `TableIterator`, `MergeIterator` and `ConcatIterator`.
+#[enum_dispatch]
+pub trait AgateIterator {
```
Connor1996, line 8: how about `AgateIterator` -> `Iterator` and `Iterator` -> `TableIterator`?

src/table/merge_iterator.rs
BusyJay: What I'm expected is
```
pub enum Iterators {
    MergeIterator(Box<MergeIterator>),
    TableIterator(TableIterator),
    #[cfg(test)]
    VecIterator(Box<tests::VecIterator>),
}
```.

On the other hand, is It possible to define `MergeIterator` as
```
struct IteratorNode<I> {
    valid: bool,
    key: BytesMut,
    iter: I,
}

pub struct MergeIterator<L, R> {
    left: IteratorNode<L>,
    right: IteratorNode<R>,
    is_left_small: bool,
    reverse: bool,
    current_key: BytesMut,
}
```?

Connor1996 (commented):
most LGTM

src/table/merge_iterator.rs
skyzh: For the former one, I'll have a try. It seems that Box<MergeIterator> doesn't implement `AgateIterator`, not sure if this would work.

For the latter one, that's exactly what I've tried before. There are several issues with this.
* First, we cannot determines type of iterator at compile-time. Our reference implementation would construct one iterator recursively from multiple iterators, when iterating multiple SSTs. It's impossible to know how many SSTs to iterator at compile time.
* Using such implementation make the merge logic very complex (as left and right are treated as different types).

src/iterator_trait.rs
skyzh: Seems that I would confuse this `Iterator` with Rust `Iterator`. What about `TableIterator` and `AgateIterator`?

src/table/merge_iterator.rs
BusyJay: From what I can see, the iterator is always be constructed from a slice of iterator? Why not store them as slice in place? So we don't need allocations and probably get better performance as it's the same direction as prefetch.

skyzh: Yes, I agree that iterators can be always stored and used as slice. But this makes the merge iterating logic very complex. We may have to build a heap to store the elements from each iterator before serving them to developers. I suggest trying this later and just keep it the same as reference implementation, by converting a slice of iterators into a binary tree of merge iterators.

src/iterator_trait.rs
Connor1996: okay

skyzh: `Iterator` -> `TableRefIterator`, and `TableRefIterator<Arc<Table>>` is now `TableIterator`

Connor1996 (approved):
LGTM

BusyJay (approved):

src/iterator_trait.rs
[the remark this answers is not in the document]
skyzh: Thanks! I'll have a try later.

PR #31, opened 2020-10-28
Fix clippy warnings and errors

AngleNet:
Fix warnings and errors while clippy

### DIFF
--- a/benches/bench_table.rs
+++ b/benches/bench_table.rs
@@ -1,5 +1,6 @@
 mod common;
 
+use agatedb::ChecksumVerificationMode::NoVerification;
 use std::ops::{Deref, DerefMut};
 
 use agatedb::{Table, TableBuilder, TableOptions, Value};
@@ -26,6 +27,7 @@ fn bench_table_builder(c: &mut Criterion) {
             block_size: 4 * 1024,
             bloom_false_positive: 0.01,
             table_size: 5 << 20,
+            checksum_mode: NoVerification,
         };
 
         b.iter(|| {
@@ -69,6 +71,7 @@ fn get_table_for_benchmark(count: usize) -> TableGuard {
         block_size: 4 * 1024,
         bloom_false_positive: 0.01,
         table_size: 0,
+        checksum_mode: NoVerification,
     };
 
     let mut builder = TableBuilder::new(opts.clone());
@@ -103,6 +106,7 @@ fn bench_table(c: &mut Criterion) {
         block_size: 4 * 1024,
         bloom_false_positive: 0.01,
         table_size: 0,
+        checksum_mode: NoVerification,
     };
 
     c.bench_function("table read and build", |b| {
--- a/benches/bench_table.rs
+++ b/benches/bench_table.rs
@@ -32,8 +32,8 @@ fn bench_table_builder(c: &mut Criterion) {
 
         b.iter(|| {
             let mut builder = TableBuilder::new(opt.clone());
-            for j in 0..KEY_COUNT {
-                builder.add(&key_list[j], vs.clone(), 0);
+            for k in key_list.iter().take(KEY_COUNT) {
+                builder.add(k, vs.clone(), 0);
             }
             builder.finish()
         });
--- a/benches/bench_table.rs
+++ b/benches/bench_table.rs
@@ -1,5 +1,7 @@
 mod common;
 
+use std::ops::{Deref, DerefMut};
+
 use agatedb::{Table, TableBuilder, TableOptions, Value};
 use bytes::Bytes;
 use common::rand_value;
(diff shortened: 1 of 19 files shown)


AngleNet:
HI, all. Is contributions welcomed for this repo ? I'd like to help ! I fixed some errors.

BusyJay:
> HI, all. Is contributions welcomed for this repo ?

Of course yes! Thanks for your contribution! Can you also add the clippy lint to CI check?

AngleNet:
Add clippy lint in GitHub workflows ?

BusyJay:
Yes.

AngleNet:
Already added the lint. By the way, Why you start to write this key-value store from scratch ? Why no rocksdb ? Could you clear this for me ? Thank you !!!

BusyJay:
Rocksdb is too complicated and general. This project is about to act quickly and bring aggressive optimizations for TiKV only. Writing it in Rust can bring memory safety and be easier integrated into TiKV. This is an experimental engine for TiKV, we will still keep using rocksdb.

AngleNet:
Got it! I looked through the code base. Is there any draft docs over the design for now ? I see that there are only some development plans to implement the core features of a classical key-value store.  Or it's planed to implement the main features for a key-value store firstly and introduce optimizations later ?

skyzh (approved):
Thanks for your contribution!

src/table/builder.rs
```diff
@@ -242,8 +240,8 @@ mod tests {
 
         let idx = table.inner.read_table_index().unwrap();
 
-        for i in 0..idx.offsets.len() {
-            assert_eq!(block_first_keys[i], idx.offsets[i].key);
+        for (i, k) in block_first_keys.iter().enumerate().take(idx.offsets.len()) {
```
BusyJay, line 243: What if `block_first_keys.len() < idx.offsets.len()`?

src/table/iterator.rs
```diff
@@ -321,10 +316,7 @@ impl<T: AsRef<TableInner>> Iterator<T> {
         let idx = util::search(self.table.as_ref().offsets_length(), |idx| {
             use std::cmp::Ordering::*;
             let block_offset = self.table.as_ref().offsets(idx).unwrap();
-            match COMPARATOR.compare_key(&block_offset.key, &key) {
-                Greater => true,
-                _ => false,
-            }
+            !matches!(COMPARATOR.compare_key(&block_offset.key, &key), Greater)
```
BusyJay, line 319: Redundant `!`.

skyzh:
> Got it! I looked through the code base. Is there any draft docs over the design for now ? I see that there are only some development plans to implement the core features of a classical key-value store. Or it's planed to implement the main features for a key-value store firstly and introduce optimizations later ?

We have updated our roadmap in README.md. Thanks for your interest.

src/table/iterator.rs
AngleNet: Oh. Yes

src/table/builder.rs
AngleNet: Will fix it!

AngleNet: @BusyJay  I think it should always be equal. If we didn't get that right. Then the test should fail. Is it correct ?

skyzh: I have added real unit tests last week. You may have to rebase your work on latest master so as to see if this change really works.

skyzh: I prefer preserving how original code works, by iterating only on offsets.

AngleNet: However, the clippy complain about the original code. I'm checking it right now.

BusyJay: > I think it should always be equal. If we didn't get that right. Then the test should fail. Is it correct ?

Yes. However after changing to iterator, even they are not equal, the test will not fail. You can either add an explicit assert on length or just dismiss the lint.

AngleNet: I prefer to add the assert of length.

BusyJay: If they have the same length, then `take(n)` is unnecessary, right? Perhaps you can write `block_first_keys..iter().zip(&idx.offsets)`.

src/table/tests.rs
```diff
@@ -84,7 +84,7 @@ fn test_table_iterator() {
 
 #[test]
 fn test_seek_to_first() {
-    for n in vec![99, 100, 101, 199, 200, 250, 9999, 10000] {
+    for &n in &[99, 100, 101, 199, 200, 250, 9999, 10000] {
```
BusyJay, line 87: In TiKV, we usually prefer `*` instead of `&`.

src/entry.rs
```diff
@@ -9,6 +10,55 @@ pub struct Entry {
     pub meta: u8,
 }
 
+fn varint_len(l: usize) -> usize {
```
BusyJay, line 13: These seem to be deleted.

benches/bench_table.rs
```diff
@@ -28,8 +28,8 @@ fn bench_table_builder(c: &mut Criterion) {
 
         b.iter(|| {
             let mut builder = TableBuilder::new(opt.clone());
-            for j in 0..KEY_COUNT {
-                builder.add(&key_list[j], vs.clone(), 0);
+            for k in key_list.iter().take(KEY_COUNT) {
```
BusyJay, line 31: How about `&key_list[..KEY_COUNT]`?

skyzh:
Thanks for your work! Would you please rebase or merge latest master onto this branch? (Or you may open a new PR for clippy). It seems that the git diff contains content from previous commits, and it's hard to review what has really changed.

AngleNet:
> Thanks for your work! Would you please rebase or merge latest master onto this branch? (Or you may open a new PR for clippy). It seems that the git diff contains content from previous commits, and it's hard to review what has really changed.

Of course! I would open another PR for this.

PR #37, opened 2020-11-03
agate: refactor for memtable impl

skyzh:
This PR adds prototypes of memtable / wal implementation. At the same time, `MemTable` is refactored to use one Wal for each memtable.


skyzh:
@BusyJay @Connor1996 This PR only contains some prototypes and is relatively easy to review, and yet is critical for implementing new functionalities. I suggest taking a look at this patch first to help me get this project move on. Thanks!

Currently, we have many PRs waiting for reviewing. Some of them are small patches and add small features, and some of them are about implementing new structures to make this project move on and complete. I have added priority labels for those PRs that is on critical path and may block me from moving ahead. For other PRs, I think it's okay to review them later.

src/db.rs
```diff
@@ -25,14 +32,40 @@ impl Agate {
         }
         unimplemented!()
     }
+    */
 }
 
-#[derive(Default)]
+#[derive(Clone)]
 pub struct AgateOptions {
```
Connor1996, line 39: Move to a separate file?

src/db.rs
```diff
@@ -41,41 +74,129 @@ impl AgateOptions {
...
+    fn fix_options(&mut self) -> Result<()> {
+        if self.in_memory {
+            // TODO: find a way to check if path is set, if set, then panic with ConfigError
+            self.sync_writes = false;
+        }
+
+        Ok(())
+    }
+
     pub fn open<P: AsRef<Path>>(&mut self, path: P) -> Result<Agate> {
```
Connor1996, line 116: can we move it to `struct Agate`?

src/db.rs
```diff
@@ -41,41 +74,129 @@ impl AgateOptions {
         self
     }
 
-    pub fn wal_path<P: Into<PathBuf>>(&mut self, p: P) -> &mut AgateOptions {
-        self.wal_path = Some(p.into());
+    pub fn path<P: Into<PathBuf>>(&mut self, p: P) -> &mut AgateOptions {
```
Connor1996, line 77: make it take `self` and return `Self`, so we can write like this 
```
let opts = AgateOptions::Default()
      .path(xxx)
      .num_memtables(xxx);
```

Also, I'm not sure if these setters are necessary. Maybe make all the fields public is enough?

src/memtable.rs
```diff
@@ -1,27 +1,70 @@
...
-                return res;
-            }
+pub struct MemTable {
+    pub(crate) skl: Skiplist<Comparator>,
+    opt: AgateOptions,
+    core: RwLock<MemTableCore>,
+}
+
+impl MemTable {
+    pub fn new(skl: Skiplist<Comparator>, wal: Option<Wal>, opt: AgateOptions) -> Self {
```
Connor1996, line 32: Not sure how it is used. But I think `skl` and `wal` should be created in `MemTable` instead of passing from outside.

src/memtable.rs
```diff
@@ -31,32 +74,34 @@ impl Drop for MemTableView {
...
 
-impl MemTable {
-    pub fn with_capacity(table_size: u32, max_count: usize) -> MemTable {
-        let c = Flsc::new(8);
-        MemTable {
-            mutable: Skiplist::with_capacity(c, table_size),
-            immutable: VecDeque::with_capacity(max_count - 1),
-        }
+impl MemTables {
+    pub(crate) fn new(mutable: MemTable, immutable: VecDeque<MemTable>) -> Self {
```
Connor1996, line 83: I prefer the previous `with_capacity` instead of new with `mutable` and `immutable` from outside

src/memtable.rs
skyzh: WAL name is decided and generated by database instance, so it is passed from outside. And for skiplist, I think we could move it inside memtable.

src/db.rs
skyzh: Implementing `open` in `XXXOptions` seems to be a convention. e.g. `OpenOptions` of `std::File`.

src/db.rs
skyzh: I'm not sure if this change is necessary. We could already write `AgateOptions::Default().path(xxx).open(xxx)` with current implementation. FYI, https://github.com/tikv/agatedb/blob/develop/src/db.rs#L317

src/memtable.rs
skyzh: The "open memtable" logic is implemented in `db.rs`, the database engine. Therefore, it's necessary to provide an interface to construct `Memtables` from a single `Memtable`.

src/util.rs
```diff
@@ -1,9 +1,15 @@
 pub mod binary;
 
+use crate::Result;
+pub use skiplist::FixedLengthSuffixComparator as Comparator;
 pub use skiplist::{FixedLengthSuffixComparator, KeyComparator};
 use std::{cmp, ptr};
 
-pub static COMPARATOR: FixedLengthSuffixComparator = FixedLengthSuffixComparator::new(8);
+pub static COMPARATOR: FixedLengthSuffixComparator = make_comparator();
```
BusyJay, line 8: What's the difference between `Comparator::new(8)`?

src/util.rs
```diff
@@ -56,3 +62,15 @@ where
     }
     i
 }
+
+/// Fill a file with 0.
+pub fn fill_file(file: &mut impl std::io::Write, mut size: u64) -> Result<()> {
```
BusyJay, line 67: Why not just `file.set_len(size)`?

src/db.rs
```diff
@@ -1,22 +1,32 @@
...
+
 use std::fs;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use std::sync::RwLock;
 
 pub struct Core {
-    wal: Wal,
-    memtable: MemTable,
+    mt: RwLock<MemTables>,
```
BusyJay, line 16: How about move the lock into `MemTables`? And `Mutex` is probably a better choice.

src/db.rs
Connor1996: Both badger and rocksdb follow the way of `DB::open(opt)`. If implementing in this way, it may cost more effect to integrate with engine trait with tikv.

src/db.rs
Connor1996: okay. Apart from that, are these setters necessary? Maybe make all the fields public is enough?

skyzh: In fact, I also prefer making all fields public (and check if they're valid when opening) cc @BusyJay

src/db.rs
skyzh: This `RwLock` guards list of `MemTables` (e.g. immutable list). Using `RwLock` ensures that we could parallel fetch data from memtables, while blocking reads only when we're flushing mutable table to immutable one.

And for moving that lock into `MemTables`, I think that is OK. Meanwhile, I would also like to know, is it our design principle to hide locks in structs, and make all functions callable with immutable reference in structs like `MemTable`, `Agate`, `LevelController`, etc.?

src/util.rs
skyzh: No difference, just to make the interface unified. What about reverting this change?

src/util.rs
skyzh: I'll try

src/db.rs
BusyJay: > If we're using Mutex, we may have to hide all memtables behind an Arc to reduce lock contention when parallel read data. Is it right?

Why? I think the time it consumes is just creating `MemTableView`, which should be very quick. `Mutex` usually has a better performance than `RwLock`, you can bench and see if it's an improvement.

All functions of `Agate` should should take immutable reference and `Agate` should be `Sync + Send`.

skyzh: Yes, thanks for pointing that out. I'll use `Mutex` for `MemTables`.

src/util.rs
skyzh: This function has been removed for now. We will directly call `set_len` on `File`, instead of creating this new function to handle that.

skyzh:
@Connor1996 PTAL, thanks!

src/error.rs
```diff
@@ -23,6 +24,12 @@ pub enum Error {
     VarDecode(&'static str),
     #[error("{0}")]
     TableRead(String),
+    #[error("Database Closed")]
+    DBClosed,
+    #[error("Lock Poison")]
+    PoisonError(String),
```
BusyJay, line 30: I suggest to panic. `Poison` is a hint of bug, we don't know what can go wrong when bug happen.

src/entry.rs
```diff
@@ -6,7 +6,25 @@ const VALUE_POINTER: u8 = 1 << 1;
...
     pub key: Bytes,
     pub value: Bytes,
-    pub meta: u8,
+    pub(crate) meta: u8,
+    pub user_meta: u8,
+    pub expires_at: u64,
+    pub(crate) version: u64,
+}
+
+pub struct EntryRef<'a> {
```
BusyJay, line 15: Why not just use `&Entry`?

src/db/opt.rs
```diff
@@ -0,0 +1,54 @@
+use super::*;
+
+#[derive(Clone)]
+pub struct AgateOptions {
+    pub path: PathBuf,
+    // TODO: docs
+    pub in_memory: bool,
+    pub sync_writes: bool,
+
+    pub create_if_not_exists: bool,
```
BusyJay, line 10: This can be removed now.

src/db/opt.rs
```diff
@@ -0,0 +1,54 @@
...
+    pub num_memtables: usize,
+    pub mem_table_size: u64,
```
BusyJay, line 12: How about using the same name as badger?

src/entry.rs
skyzh: `Entry` uses `Bytes` internally. But sometimes we could only get `&[u8]` (if not copy).

skyzh: FYI, https://github.com/tikv/agatedb/blob/develop/src/value.rs#L185

The key and value is a `Vec<u8>` (or you may think of it as `BytesMut`), which will only be valid for one iteration. This way, we let our user to decide whether to copy it into a `Bytes`.

BusyJay: Got it.

src/db/opt.rs
skyzh: I've added all supported options in develop branch. Now they have the same naming convention as badger.

src/db.rs
Connor1996: PTAL again

skyzh: Yes, I've already moved `open` from `AgateOptions` into `Agate`

Connor1996 (approved):
LGTM

src/db/opt.rs
```diff
@@ -0,0 +1,78 @@
+use super::*;
+
+#[derive(Clone)]
+pub struct AgateOptions {
+    pub path: PathBuf,
```
BusyJay, line 5: How about also using `dir`?

src/db/opt.rs
```diff
@@ -0,0 +1,78 @@
...
+    // TODO: docs
+    pub in_memory: bool,
+    pub sync_writes: bool,
+
+    // Memtable options
+    pub mem_table_size: u64,
+    pub base_table_size: u64,
+    pub base_level_size: u64,
+    pub level_size_multiplier: usize,
+    pub table_size_multiplier: usize,
```
BusyJay, line 16: What is `table_size_multiplier`?

src/db.rs
```diff
@@ -25,57 +35,80 @@ impl Agate {
...
+        unimplemented!()
+    }
+}
+
+impl Agate {
+    pub fn get(&self, key: &[u8]) -> Result<Value> {
+        self.core.get(key)
+    }
+
+    pub fn write_to_lsm(&self, request: Request) -> Result<()> {
```
BusyJay, line 93: I think it should be `pub(crate)`.

src/memtable.rs
```diff
@@ -1,27 +1,70 @@
...
-    pub fn get(&self, key: &[u8]) -> Option<&Bytes> {
-        for i in 0..self.len {
-            let res = self.tables[i].get(key);
-            if res.is_some() {
-                return res;
-            }
+pub struct MemTable {
+    pub(crate) skl: Skiplist<Comparator>,
+    opt: AgateOptions,
+    core: RwLock<MemTableCore>,
```
BusyJay, line 28: Why use `RwLock`?

src/wal.rs
```diff
@@ -59,7 +69,7 @@ impl Header {
     }
 
     /// Decode header from bytes
-    pub fn decode(&mut self, bytes: &mut Bytes) -> Result<usize> {
+    pub fn decode(&mut self, bytes: &[u8]) -> Result<usize> {
```
BusyJay, line 72: This should break implementation of iterator, which depends on `decode` mutating `Bytes`.

src/db/opt.rs
skyzh: This option is defined at https://github.com/dgraph-io/badger/blob/master/options.go#L58 , which seems to be used when calculating target file size of a LSM level.

src/memtable.rs
skyzh: This `RwLock` protects `Wal`. It seems that when database is running, we only write to `Wal`, and we don't need to read from it when reading from memtable. Perhaps `Mutex` is a better choice, I'll change that.

src/db/opt.rs
BusyJay: Turns out I'm referencing an outdated version. 🤣

src/wal.rs
skyzh: We could always convert `&Bytes` to `&[u8]` without overhead. And when we're decoding header from `EntryRef`, we had to use `&[u8]`. So I think this change is reasonable.

BusyJay: I'm not objecting the changes. I'm suggesting the change introduces a bug at https://github.com/tikv/agatedb/blob/afb59d2b2cf294408ceb0b186d08b6a19dfb662c/src/table/iterator.rs#L123-L139 that needs to be fixed.

skyzh: That's not the same header. I think this naming might be confusing. This is the header for WAL, and the one in iterator is the SST header. I think we could later unify these interfaces :)

skyzh: What about using `TableHeader` for SST, and `WalHeader` for write-ahead log / vlog? Meanwhile, we could change all `decode` to use `&[u8]` as parameter to reduce the need of using `slice` on `Bytes` (which introduces overhead of increasing the reference counter)

BusyJay: LGTM, but better another PR.

skyzh: Alright. I've reverted this change.

BusyJay (approved):

src/db/opt.rs
skyzh: I have forked badger@45bca18 at https://github.com/skyzh/badger . Seems that badger is under constant change, with new things added and some logic changed. We could always use my fork as reference implementation.

PR #38, opened 2020-11-04
table: add extra tests

skyzh:
This PR adds extra tests to ensure behavior of the iterators. It also fixes https://github.com/tikv/agatedb/issues/35.

### DIFF
--- a/benches/bench_table.rs
+++ b/benches/bench_table.rs
@@ -1,5 +1,7 @@
 mod common;
 
+use std::ops::{Deref, DerefMut};
+
 use agatedb::{Table, TableBuilder, TableOptions, Value};
 use bytes::Bytes;
 use common::rand_value;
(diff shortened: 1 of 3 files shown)


src/table/tests.rs
```diff
@@ -344,3 +370,67 @@ fn test_table_big_values() {
...
+
+#[test]
+fn test_iterator_out_of_bound_reverse() {
+    let opts = get_test_table_options();
+    let table = build_test_table(b"key", 1000, opts);
+    let mut it = table.new_iterator(ITERATOR_REVERSED);
+    it.seek_to_first();
+    assert!(it.error().is_none());
+    it.next();
+    assert!(matches!(it.error(), Some(IteratorError::EOF)));
```
BusyJay, line 428: Can `IteratorError` derive `PartialEq`? So you can use `assert_eq`, which produces more readable error when fail.

skyzh: I've added `PartialEq`, and now in tests, we use `== Some` instead of `matches!`. But in places other than test, I retained the original implementation (`matches!(...)`). Do we need to also change how it works in main program?

benches/bench_table.rs
```diff
@@ -36,7 +38,30 @@ fn bench_table_builder(c: &mut Criterion) {
     });
 }
 
-fn get_table_for_benchmark(count: usize) -> Table {
+/// TableGuard saves Table and TempDir, so as to ensure
+/// temporary directory is removed after table is closed.
+/// According to Rust RFC, the drop order is first `table` then
+/// `tmp_dir`.
+pub struct TableGuard {
```
Connor1996, line 45: why implement it in two places?

Connor1996 (commented):
rest LGTM

benches/bench_table.rs
skyzh: This struct is only used in test / bench. I think I could find a better way to use them in only test / bench while only implementing it once.

BusyJay (approved):

benches/bench_table.rs
skyzh: Seems there is no way to export TableGuard only on bench. FYI, https://users.rust-lang.org/t/conditional-compilation-for-benchmarks/47227/2

> External `tests/` and `benches/` always get a non-test version of the library.

Therefore, the only way seems to be creating a `TableGuard` that will be embedded into release binary. What's your opinion?

Connor1996: alright

Connor1996 (approved):

PR #47, opened 2020-11-10
[dev] levels: add L0 flush

skyzh:


skyzh:
From this PR we begin to add multi-thread related functionalities. I plan to use a unified thread pool `yatp` in every agate db instance.

Currently, `Agate` is simply a wrapper of `Core`. After adding the thread pool, `Agate` includes both an `Arc<Core>` and a thread pool. Functions running in thread pool will be able to own an `Arc<Core>`. And `Agate` is now non-clonable.

PR #54, opened 2020-11-24
[dev] levels: finish compaction

skyzh:


skyzh:
Get key is very slow due to `ConcatIterator` is not implemented. Tests would timeout very easily.

skyzh:
Get key optimization has been implemented.

PR #55, opened 2020-11-24
agate: add write-ahead log

skyzh:

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -172,7 +172,7 @@ impl EntryReader {
         self.header.decode(reader)?;
         if self.header.key_len > (1 << 16) {
             return Err(Error::LogRead(
-                "key length must be lower than 1 << 16".to_string(),
+                "key length must not be larger than 1 << 16".to_string(),
             ));
         }
         self.key.resize(self.header.key_len as usize, 0);
--- a/src/value.rs
+++ b/src/value.rs
@@ -175,6 +175,7 @@ impl EntryReader {
                 "key length must not be larger than 1 << 16".to_string(),
             ));
         }
+        // TODO: resize key and value without initialization
         self.key.resize(self.header.key_len as usize, 0);
         reader.read_exact(&mut self.key)?;
         self.value.resize(self.header.value_len as usize, 0);
--- a/src/value.rs
+++ b/src/value.rs
@@ -3,8 +3,7 @@ use crate::entry::EntryRef;
 use crate::wal::Header;
 use crate::Result;
 use bytes::{BufMut, Bytes, BytesMut};
-use std::io::BufReader;
-use std::io::{Read, Seek};
+use std::io::{Cursor, Read, Seek};
 use std::mem::MaybeUninit;
 
 pub const VALUE_DELETE: u8 = 1 << 0;
@@ -169,15 +168,8 @@ impl EntryReader {
     }
 
     /// Entry returns header, key and value.
-    pub fn entry<R: Read + Seek>(&mut self, reader: &mut BufReader<R>) -> Result<EntryRef> {
-        // first, read MAX_HEADER_SIZE bytes to buf
-        let size = reader.read(&mut self.buf)?;
-        // then, try decode the header
-        let decoded_size = self.header.decode(&self.buf[0..size as usize])?;
-        // now, we seek back to where the header ends
-        reader.seek(std::io::SeekFrom::Current(
-            decoded_size as i64 - size as i64,
-        ))?;
+    pub fn entry(&mut self, reader: &mut Cursor<&[u8]>) -> Result<EntryRef> {
+        self.header.decode(reader)?;
         self.key.resize(self.header.key_len as usize, 0);
         reader.read_exact(&mut self.key)?;
         self.value.resize(self.header.value_len as usize, 0);
--- a/src/value.rs
+++ b/src/value.rs
@@ -169,7 +169,26 @@ impl EntryReader {
     }
 
     /// Entry returns header, key and value.
-    pub fn entry<R: Read + Seek>(&mut self, _reader: &mut BufReader<R>) -> Result<EntryRef> {
-        unimplemented!()
+    pub fn entry<R: Read + Seek>(&mut self, reader: &mut BufReader<R>) -> Result<EntryRef> {
+        // first, read MAX_HEADER_SIZE bytes to buf
+        let size = reader.read(&mut self.buf)?;
+        // then, try decode the header
+        let decoded_size = self.header.decode(&self.buf[0..size as usize])?;
+        // now, we seek back to where the header ends
+        reader.seek(std::io::SeekFrom::Current(
+            decoded_size as i64 - size as i64,
+        ))?;
+        self.key.resize(self.header.key_len as usize, 0);
+        reader.read_exact(&mut self.key)?;
+        self.value.resize(self.header.value_len as usize, 0);
+        reader.read_exact(&mut self.value)?;
+        Ok(EntryRef {
+            key: &self.key,
+            value: &self.value,
+            meta: self.header.meta,
+            user_meta: self.header.user_meta,
+            expires_at: self.header.expires_at,
+            version: 0,
+        })
     }
 }
--- a/src/value.rs
+++ b/src/value.rs
@@ -1,7 +1,7 @@
 use crate::entry::Entry;
 use crate::entry::EntryRef;
 use crate::wal::Header;
-use crate::Result;
+use crate::{Error, Result};
 use bytes::{BufMut, Bytes, BytesMut};
 use std::io::{Cursor, Read};
 use std::mem::MaybeUninit;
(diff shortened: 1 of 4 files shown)


src/value.rs
```diff
@@ -169,7 +169,26 @@ impl EntryReader {
     }
 
     /// Entry returns header, key and value.
-    pub fn entry<R: Read + Seek>(&mut self, _reader: &mut BufReader<R>) -> Result<EntryRef> {
-        unimplemented!()
+    pub fn entry<R: Read + Seek>(&mut self, reader: &mut BufReader<R>) -> Result<EntryRef> {
+        // first, read MAX_HEADER_SIZE bytes to buf
+        let size = reader.read(&mut self.buf)?;
```
BusyJay, line 174: What if the reader reads an part of the data?

skyzh: What about using `read_exact` here and ignore EOF here? The only possibility that we read less than `MAX_HEADER_SIZE` is when it comes to end of file. If so, we decode the entry normally. For other errors, just panic.

src/wal.rs
```diff
@@ -100,69 +104,242 @@ pub struct Wal {
...
+            size: mmap_file.len() as u32,
+            mmap_file,
+            opts,
+            write_at: 0,
+            // TODO: current implementation doesn't have keyID and baseIV header
+            buf: BytesMut::new(),
+        };
+
+        if bootstrap {
+            wal.bootstrap()?;
```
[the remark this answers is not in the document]
skyzh: For deleting files in agatedb, I'll introduce it later along with database manifest implementation. Currently, we just retain all files inside it.

src/value.rs
BusyJay: Is it possible to know the length ahead of calling the function?

skyzh: Value log or WAL is simply a sequence of records, where each record begins with a variable-size header and variable-size key + value. So it's impossible to know how large is header before actually reading from that. Meanwhile, we could also use an alternative encoding different from badger (e.g. fixed-size header)

BusyJay: I have an idea. If you implement `Buf` for a `BufReader` using a wrapper, you can reuse `prost::decode_length_delimiter` and `prost::encode_length_delimiter` to decode the header.

skyzh: Now we use `prost::encoding::decode_varint` for this operation (which takes a `Buf`). In next few PRs, I would phase out `binary.rs` and use utilities provided by prost.

src/wal.rs
```diff
@@ -100,69 +94,244 @@ pub struct Wal {
 }
 
 impl Wal {
-    pub fn open(_path: PathBuf, _opts: AgateOptions) -> Result<Wal> {
-        unimplemented!()
+    /// open or create a WAL from options
+    pub fn open(path: PathBuf, opts: AgateOptions) -> Result<Wal> {
+        let (file, bootstrap) = if path.exists() {
```
BusyJay, line 99: Seems need to check the file length.

skyzh: This simple WAL implementation do not have vlog header (which only includes encryption-related information). Thus we can directly start reading from it.

src/value.rs
```diff
@@ -169,7 +168,24 @@ impl EntryReader {
...
-    pub fn entry<R: Read + Seek>(&mut self, _reader: &mut BufReader<R>) -> Result<EntryRef> {
-        unimplemented!()
+    pub fn entry(&mut self, reader: &mut Cursor<&[u8]>) -> Result<EntryRef> {
+        self.header.decode(reader)?;
+        if self.header.key_len > (1 << 16) {
+            return Err(Error::LogRead(
+                "key length must be lower than 1 << 16".to_string(),
+            ));
+        }
+        self.key.resize(self.header.key_len as usize, 0);
```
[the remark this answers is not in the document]
skyzh: I've added a TODO here.

src/wal.rs
BusyJay: What if the file is smaller than the given size?

src/wal.rs
```diff
@@ -100,69 +94,246 @@ pub struct Wal {
...
+        self.mmap_file.flush()?;
+        Ok(())
     }
 
     pub fn zero_next_entry(&mut self) -> Result<()> {
-        unimplemented!()
+        let range =
+            &mut self.mmap_file[self.write_at as usize..self.write_at as usize + MAX_HEADER_SIZE];
+        // this code could be optimized by compiler to write 8 bytes a time
+        for x in range {
```
BusyJay, line 164: How about `ptr::write_bytes(range.get_mut_ptr(), 0, range.len())`;

src/wal.rs
```diff
@@ -100,69 +94,246 @@ pub struct Wal {
...
-        unimplemented!()
+    /// Truncate WAL
+    pub fn truncate(&mut self, end: u64) -> Result<()> {
+        // TODO: check read only
+        let metadata = self.file.metadata()?;
+        if metadata.len() == end {
+            return Ok(());
+        }
+        self.size = end as u32;
+        self.file.set_len(end)?;
```
BusyJay, line 237: Should it be `sync`?

src/wal.rs
```diff
@@ -100,69 +94,246 @@ pub struct Wal {
...
+        match entry {
+            Ok(entry) => {
+                if entry.is_zero() {
+                    return None;
+                }
+                // TODO: process transaction-related metadata
+                Some(Ok(entry))
+            }
+            Err(Error::Io(err)) => {
+                if err.kind() == ErrorKind::UnexpectedEof {
```
BusyJay, line 315: What if it reads a truncated entry?

skyzh: I have added tests for reading truncated entry.

There're 3 kinds of error when we read a truncated entry.

* If header is less than 2 bytes, we could not read meta. There will be `VarDecode` error.
* If varint key length and value length is truncated, prost will return `Decode` error.
* If key / value is truncated, `Cursor` will return `UnexpectedEof`.

BusyJay: Then how to tell if it's truncated or just reach the end of the file?

Another question is why return `Option<Result>` instead of `Result<Option>`?

skyzh: `None` means a recoverable error (e.g. truncate, decode error), and `Some(Err)` refers to an unexpected error. Maybe it's better to exchange them.

And for the truncate issue, I'm a little bit confused about in what state a file is considered `truncate`. I'll look into this soon.

skyzh: From https://github.com/dgraph-io/badger/blob/master/memtable.go#L475 , we do not need to tell whether it's truncated or reached end of file. In both situation, we simply stop reading from the log file.

I've modified current implementation. Now it will:
* return `Ok(None)` if we reached first corrupted entry, and we could stop iteration
* return `Ok(Some(entry))` if we read a new entry
* return `Err` if the error is not recoverable

src/wal.rs
skyzh: It seems that we do not have a pre-defined size for WAL. We just simply read from start to the end.

BusyJay (approved):

src/wal.rs
```diff
@@ -100,69 +97,256 @@ pub struct Wal {
...
+        unsafe {
+            std::ptr::write_bytes(range.as_mut_ptr(), 0, range.len());
+        }
+        Ok(())
     }
 
-    fn encode_entry(_buf: &mut BytesMut, _entry: &Entry) {
-        unimplemented!()
+    /// Encode entry to buffer
+    pub(crate) fn encode_entry(mut buf: &mut BytesMut, entry: &Entry) -> usize {
```
Connor1996, line 173: add format description here

src/wal.rs
```diff
@@ -100,69 +97,256 @@ pub struct Wal {
...
+            // ignore prost varint decode error
+            Err(Error::Decode(_)) => Ok(None),
+            // ignore custom decode error (e.g. header <= 2)
+            Err(Error::VarDecode(_)) => Ok(None),
+            // ignore file length < key, value size
+            Err(Error::Io(err)) => {
+                if err.kind() == ErrorKind::UnexpectedEof {
+                    Ok(None)
+                } else {
+                    return Err(Error::Io(err));
```
Connor1996, line 331:
```suggestion
                   Err(Error::Io(err))
```

src/wal.rs
```diff
@@ -100,69 +97,256 @@ pub struct Wal {
...
+                false,
+            )
+        } else {
+            let file = OpenOptions::new()
+                .create_new(true)
+                .read(true)
+                .write(true)
+                .open(&path)?;
+            file.set_len(2 * opts.value_log_file_size)?;
+            file.sync_all()?;
```
Connor1996, line 118: need to sync dir too

src/wal.rs
```diff
@@ -170,6 +172,11 @@ impl Wal {
     }
 
     /// Encode entry to buffer
+    ///
+    /// The entry is encoded to a header followed by plain key and value.
```
Connor1996, line 176: What I want is like this.
截屏2020-12-09 上午11 18 59

"Header consists of ...." should be included in the comment of header encode() instead of here.

skyzh: Alright, I'll add this.

skyzh: Header format and WAL format has been added.

Connor1996 (approved):

PR #65, opened 2020-11-27
table: add concat iterator

skyzh:


src/table/concat_iterator.rs
```diff
@@ -0,0 +1,197 @@
...
+pub struct ConcatIterator {
+    cur: Option<usize>,
+    iters: Vec<Option<TableIterator>>,
+    tables: Vec<Table>,
+    opt: usize,
+}
+
+impl ConcatIterator {
+    pub fn from_tables(tables: Vec<Table>, opt: usize) -> Self {
+        let mut iters = vec![];
```
BusyJay, line 17: Should use `Vec::with_capacity()`.

src/table/merge_iterator.rs
```diff
@@ -281,47 +289,95 @@ mod tests {
...
-    fn check_sequence(mut iter: Box<Iterators>, n: usize) {
+    fn check_sequence_both(mut iter: Box<Iterators>, n: usize, reversed: bool) {
+        // test sequentially iterate
         let mut cnt = 0;
         iter.rewind();
         while iter.valid() {
-            assert_eq!(user_key(iter.key()), format!("{:012x}", cnt).as_bytes());
+            let check_cnt = if reversed { n - 1 - cnt } else { cnt };
+            assert_eq!(
+                Bytes::copy_from_slice(user_key(iter.key())),
```
BusyJay, line 299: Why clone?

skyzh: When asserting eq, `Bytes` is more readable than `&[u8]` in output, as it will show characters as-is and only escape control characters. This performance penalty is acceptable as it's only used in tests.

BusyJay: Cool, how about making it an explicit function and comment the intentions?

src/table.rs
```diff
@@ -1,6 +1,13 @@
 pub(crate) mod builder;
+pub mod concat_iterator;
 mod iterator;
-mod merge_iterator;
+pub mod merge_iterator;
+
+pub use concat_iterator::ConcatIterator;
+use iterator::{TableRefIterator, ITERATOR_NOCACHE, ITERATOR_REVERSED};
```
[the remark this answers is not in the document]
skyzh: Now all public exports are at the front, and all others are below.

src/table/merge_iterator.rs
skyzh: I have added `assert_bytes_eq`, which first cast `&[u8]` into `Bytes`, and then `assert_eq!`.

src/table/merge_iterator.rs
```diff
@@ -281,47 +289,101 @@ mod tests {
             .collect()
     }
 
-    fn check_sequence(mut iter: Box<Iterators>, n: usize) {
+    macro_rules! assert_bytes_eq {
```
BusyJay, line 292: Still missing comment explaining why use `Bytes`.

BusyJay (commented):
Rest LGTM

src/table/merge_iterator.rs
skyzh: done

BusyJay (approved):

src/table/concat_iterator.rs
```diff
@@ -0,0 +1,200 @@
...
+            self.cur = None;
+            return;
+        }
+        if self.iters[idx].is_none() {
+            self.iters[idx] = Some(self.tables[idx].new_iterator(self.opt));
+        }
+        self.cur = Some(idx);
+    }
+
+    fn iter_mut(&mut self, idx: usize) -> &mut TableIterator {
```
Connor1996, line 41: seems the `idx` is always self.cur.unwrap(), can we remove the parameter?

Connor1996 (commented):
rest LGTM

src/table/concat_iterator.rs
skyzh: `self.cur` may become `None` when the iterator points to invalid location. So currently we could not remove this.

The underlying issue is that, badger use `-1` to represent iterator beyond reverse iteration, and we always use `usize` for iterator index. Therefore, we have to represent `cur` either in `Some/None` way, or represent `-1` as `usize::max` like what we've done in `TableIterator`.

Connor1996: but now, `iter_mut` is always called when `self.cur` is some.

skyzh: Got it. I misunderstood previous comments. I'll remove the parameter from function signature.

skyzh: done

Connor1996 (approved):

PR #68, opened 2020-12-01
skiplist: add owned iterator

skyzh:
This PR adds owned iterator for `Skiplist`. The iterator will own a `Skiplist` instance.

This is useful when processing a transaction. When starting a transaction, all iterators (including memtable, SST, and pending writes) will be saved inside txn struct, and being packed into a `MergeIterator`. Hence, we must have a owned iterator instead of using an iterator referencing a skiplist.

FYI: https://github.com/dgraph-io/badger/blob/master/iterator.go#L457

Not sure if this change would introduce memory flaws.

### DIFF
--- a/skiplist/src/key.rs
+++ b/skiplist/src/key.rs
@@ -1,7 +1,7 @@
 use bytes::Bytes;
 use std::cmp::Ordering;
 
-pub trait KeyComparator {
+pub trait KeyComparator: Clone {
     fn compare_key(&self, lhs: &[u8], rhs: &[u8]) -> Ordering;
     fn same_key(&self, lhs: &[u8], rhs: &[u8]) -> bool;
 }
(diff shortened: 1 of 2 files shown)


BusyJay (approved):

Connor1996:
Sorry, I can't understand why it needs to own a skiplist.
https://github.com/dgraph-io/badger/blob/master/iterator.go#L457 seems the line number has changed?

skyzh:
> Sorry, I can't understand why it needs to own a skiplist.
> https://github.com/dgraph-io/badger/blob/master/iterator.go#L457 seems the line number has changed?

It's here now. https://github.com/dgraph-io/badger/blob/b80c792be789ef9a0c698c866a3bb4df093953b9/iterator.go#L473

Connor1996 (approved):

PR #70, opened 2020-12-01
skiplist: support get value with key

skyzh:
When we're iterating memtables, we not only need value of key (at some timestamp), but also need its timestamp to see whether it is within `read_ts`. Therefore, `get_with_key` is added to get the corresponding key.

### DIFF
--- a/skiplist/src/list.rs
+++ b/skiplist/src/list.rs
@@ -281,12 +281,20 @@ impl<C: KeyComparator> Skiplist<C> {
     }
 
     pub fn get(&self, key: &[u8]) -> Option<&Bytes> {
+        if let Some((_, value)) = self.get_with_key(key) {
+            Some(value)
+        } else {
+            None
+        }
+    }
+
+    pub fn get_with_key(&self, key: &[u8]) -> Option<(&Bytes, &Bytes)> {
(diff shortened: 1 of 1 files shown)


skiplist/src/list.rs
```diff
@@ -281,12 +281,20 @@ impl<C: KeyComparator> Skiplist<C> {
...
+    }
+
+    pub fn get_with_key(&self, key: &[u8]) -> Option<(&Bytes, &Bytes)> {
         let node = unsafe { self.find_near(key, false, true) };
         if node.is_null() {
             return None;
         }
         if self.c.same_key(&unsafe { &*node }.key, key) {
-            return unsafe { Some(&(*node).value) };
+            return Some(unsafe { (&(*node).key, &(*node).value) });
```
BusyJay, line 297: How about returning version just like what `badger` does?

skyzh: The issue is that, we could not know what part of the key is the version in `skiplist` crate. One solution is to add `get_version` in `KeyComparator` trait. What about doing this?

BusyJay: Get it.

BusyJay (approved):

Connor1996 (approved):

PR #82, opened 2020-12-09
vlog: add value log

skyzh:
vLog stores large values (generally >= 4KB). The writing process of large value is as follows:

* Upon receiving the request in the flushing thread, write the value to vLog.
* Now we have offset and file ID, and we use this as value and write it to memtable and LSM tree's WAL.
* The key-(vLog offset) pair will be processed in the same way as other key-value pairs in LSM tree.

Agate vLog is simply a WAL. There's no checksum and encryption support. We simply write data one by one into the vLog. vLog GC is not implemented and will be done later.


src/value_log.rs
```diff
@@ -0,0 +1,331 @@
...
+
+    fn sorted_fids(&self) -> Vec<u32> {
+        let core = self.core.read().unwrap();
+        let mut to_be_deleted = HashSet::new();
+        for fid in core.files_to_delete.iter().cloned() {
+            to_be_deleted.insert(fid);
+        }
+        let mut result = vec![];
+        for (fid, _) in core.files_map.iter() {
+            if to_be_deleted.get(fid).is_none() {
```
BusyJay, line 100: Why not `!to_be_deleted.contains(fid)`?

src/value_log.rs
```diff
@@ -0,0 +1,331 @@
...
+                result.push(*fid);
+            }
+        }
+        result.sort();
+        result
+    }
+
+    /// Open value log directory
+    pub fn open(&self) -> Result<()> {
```
BusyJay, line 109: Why not make it a contructor?

src/value_log.rs
```diff
@@ -0,0 +1,331 @@
...
+            let current_log_ptr = core.files_map.get(&current_log_id).unwrap().clone();
+            let mut current_log = current_log_ptr.lock().unwrap();
+            drop(core);
+            current_log.sync()?;
+        }
+        result
+    }
+
+    pub fn write_inner(&self, requests: &mut [Request]) -> Result<()> {
+        // TODO: validate writes
```
BusyJay, line 139: Seems need to check if it's in memory mode first.

src/value_log.rs
```diff
@@ -0,0 +1,331 @@
...
+                    continue;
+                }
+
+                let mut p = ValuePointer::default();
+
+                p.file_id = current_log_id;
+                p.offset = self.w_offset();
+
+                let orig_meta = entry.meta;
+                entry.meta = entry.meta & (!value::VALUE_FIN_TXN | value::VALUE_TXN);
```
BusyJay, line 200: How about `&=`?

src/value_log.rs
```diff
@@ -0,0 +1,331 @@
...
+            // TODO: handle value >= 4GB case
+            if end_offset >= current_log.size() {
+                current_log.set_len(end_offset as u64)?;
+            }
+            (&mut current_log.data()[start as usize..end_offset as usize]).clone_from_slice(buf);
+            current_log.set_size(end_offset);
+            Ok(())
+        };
+
+        let to_disk = |current_log: Arc<Mutex<Wal>>| -> Result<bool> {
```
BusyJay, line 167: Why not take `&Mutex<Wal>`?

src/value_log.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn write_inner(&self, requests: &mut [Request]) -> Result<()> {
+        // TODO: validate writes
+
+        let core = self.core.read().unwrap();
+        let mut current_log_id = core.max_fid;
+        let mut current_log = core.files_map.get(&current_log_id).unwrap().clone();
+        drop(core);
+
+        let write = |buf: &[u8], current_log: Arc<Mutex<Wal>>| -> Result<()> {
+            let mut current_log = current_log.lock().unwrap();
```
BusyJay, line 147: Can be moved to L158.

src/entry.rs
```diff
@@ -40,7 +39,15 @@ impl Entry {
     }
 
     pub fn mark_delete(&mut self) {
-        self.meta |= DELETE;
+        self.meta |= VALUE_DELETE;
```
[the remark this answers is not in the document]
skyzh: Now we define value-related bits in one place `value.rs`

```
pub const VALUE_DELETE: u8 = 1 << 0;
pub const VALUE_POINTER: u8 = 1 << 1;
pub const VALUE_DISCARD_EARLIER_VERSIONS: u8 = 1 << 2;
pub const VALUE_MERGE_ENTRY: u8 = 1 << 3;
pub const VALUE_TXN: u8 = 1 << 6;
pub const VALUE_FIN_TXN: u8 = 1 << 7;
```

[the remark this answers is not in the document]
skyzh: `DELETE` and `VALUE_POINTER` in `entry.rs` is replaced by `VALUE_DELETE` and `VALUE_POINTER` in `value.rs`

src/value_log.rs
skyzh: If `ValueLog` is created, it must be on-disk. `ValueLog::new` will return `None` if in-memory mode is enabled.

src/value_log.rs
skyzh: Now `open` is private, and the vlog directory will be opened upon `new` is called.

skyzh:
I've updated the code a little bit and make the description of this PR clearer. I'll take some time to review myself to check if the current patch is consistent with the develop branch and find potential issues. This could be done this week :)

skyzh:
This PR is ready for review.

src/wal.rs
```diff
@@ -425,3 +425,9 @@ mod tests {
         }
     }
 }
+
+impl Drop for Wal {
+    fn drop(&mut self) {
+        self.file.sync_all().expect("failed to sync WAL");
```
BusyJay, line 431: Doesn't seem to be a good idea. Users should sync explicitly if they wish.

src/value.rs
```diff
@@ -136,33 +136,50 @@ impl Value {
...
-/// `EntryReader` reads entries from `BufReader`.
+impl ValuePointer {
+    pub fn decode(&mut self, mut bytes: &[u8]) {
+        self.file_id = bytes.get_u32();
+        self.len = bytes.get_u32();
+        self.offset = bytes.get_u32();
+    }
+
+    pub fn encode(&self, buf: &mut BytesMut) {
+        buf.put_u32(self.file_id);
```
BusyJay, line 165: How about using little endian? Given that most of processors use little endians, so it will be a no-op.

src/error.rs
```diff
@@ -27,6 +29,16 @@ pub enum Error {
     DBClosed,
     #[error("Error when reading from log: {0}")]
     LogRead(String),
+    #[error("Invalid VP: {vptr:?}, kvlen {kvlen}, {range:?}")]
+    InvalidValuePointer {
+        vptr: ValuePointer,
```
BusyJay, line 34: Is it a rare error? How about Boxing the payload to reduce error struct size?

src/entry.rs
```diff
@@ -40,7 +39,15 @@ impl Entry {
...
+    }
+
+    pub fn estimate_size(&self, threshold: usize) -> usize {
+        if self.value.len() < threshold {
+            self.key.len() + self.value.len() + 2
```
BusyJay, line 47: How about adding some comments to explain what are those hard-coded numbers?

src/value.rs
skyzh: Now we use little-endian for vptr encoding.

Previously in SSTs, we used big-endian to encode length. We may later determine what endian to use in all places.

src/value_log.rs
```diff
@@ -0,0 +1,374 @@
...
+    fn file_path(&self, fid: u32) -> PathBuf {
+        vlog_file_path(&self.dir_path, fid)
+    }
+
+    fn populate_files_map(&self) -> Result<()> {
+        let dir = std::fs::read_dir(&self.dir_path)?;
+        let mut core = self.core.write().unwrap();
+        for file in dir {
+            let file = file?;
+            if let Ok(filename) = file.file_name().into_string() {
```
BusyJay, line 83: Should it return errors?

src/value_log.rs
```diff
@@ -0,0 +1,374 @@
...
+        };
+
+        // `to_disk` returns `true` if we need a new vLog.
+        let to_disk = |current_log: &RwLock<Wal>| -> Result<bool> {
+            let core = self.core.read().unwrap();
+            if self.w_offset() as u64 > self.opts.value_log_file_size
+                || core.num_entries_written > self.opts.value_log_max_entries
+            {
+                let mut current_log = current_log.write().unwrap();
+                current_log.done_writing(self.w_offset())?;
```
BusyJay, line 215: This seems can race with L181, is it safe?

skyzh: All sub-functions in this `write_inner` will not be spawned to other threads. They are all running in one thread. Things like `to_disk` and `write` are used many times throughout the vLog writing process, so they are made into two separate sub-routines.

skyzh: In fact, I really think Badger needs refactoring this part...

src/entry.rs
```diff
@@ -40,7 +39,20 @@ impl Entry {
...
 
     pub fn mark_delete(&mut self) {
-        self.meta |= DELETE;
+        self.meta |= VALUE_DELETE;
+    }
+
+    pub fn estimate_size(&self, threshold: usize) -> usize {
+        // The estimated size of an entry will be key length + value length +
+        // two bytes of metadata + expires_at.
+        const METADATA_SIZE: usize = std::mem::size_of::<u8>() * 2 + std::mem::size_of::<u64>();
```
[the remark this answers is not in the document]
skyzh: expires_at is now removed.

src/value_log.rs
```diff
@@ -0,0 +1,382 @@
...
+use crate::value::{self, Request, ValuePointer};
+use crate::wal::{Header, Wal};
+use crate::AgateOptions;
+use crate::{Error, Result};
+
+use bytes::{Bytes, BytesMut};
+use std::collections::{HashMap, HashSet};
+use std::path::{Path, PathBuf};
+use std::sync::atomic::AtomicU32;
+use std::sync::{Arc, RwLock};
```
BusyJay, line 11: How about using `parking_lot::RwLock`?

BusyJay (commented):
Rest LGTM

src/value_log.rs
skyzh: Good to have that.

BusyJay (approved):

PR #85, opened 2020-12-11
levels: prepare for implementing LSM

skyzh:
This PR adds necessary structures for implementing LSM, and added delete support for SST.


skyzh:
@BusyJay @Connor1996 PTAL if you have time~ Thanks!

src/levels/compaction.rs
```diff
@@ -0,0 +1,326 @@
...
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub struct KeyRange {
+    pub left: Bytes,
+    pub right: Bytes,
+    pub inf: bool,
+}
+
+impl KeyRange {
+    pub fn is_empty(&self) -> bool {
+        if self.inf {
```
BusyJay, line 23: Why not write it in a line?

src/levels/compaction.rs
```diff
@@ -0,0 +1,326 @@
...
+            *self = range.clone();
+            return;
+        }
+        if range.left.len() == 0
+            || COMPARATOR.compare_key(&range.left, &self.left) == std::cmp::Ordering::Less
+        {
+            self.left = range.left.clone();
+        }
+        if range.right.len() == 0
+            || COMPARATOR.compare_key(&range.left, &self.left) == std::cmp::Ordering::Greater
```
BusyJay, line 58: You mean `range.right` and `self.right`?

src/levels/compaction.rs
```diff
@@ -0,0 +1,326 @@
...
+        }
+    }
+
+    /// Extend current key range with another
+    pub fn extend(&mut self, range: &Self) {
+        if self.is_empty() {
+            *self = range.clone();
+            return;
+        }
+        if range.left.len() == 0
```
BusyJay, line 52: It should be `self.left.is_empty()`.

src/levels/compaction.rs
```diff
@@ -0,0 +1,326 @@
...
+            return false;
+        }
+        if COMPARATOR.compare_key(&self.right, &dst.left) == std::cmp::Ordering::Less {
+            return false;
+        }
+        true
+    }
+}
+
+impl Default for KeyRange {
```
BusyJay, line 87: Is it necessary?

src/levels/compaction.rs
```diff
@@ -0,0 +1,326 @@
...
+pub struct LevelCompactStatus {
+    pub ranges: Vec<KeyRange>,
+    pub del_size: u64,
+}
+
+impl LevelCompactStatus {
+    pub fn remove(&mut self, dst: &KeyRange) -> bool {
+        let prev_ranges_len = self.ranges.len();
+        // TODO: remove in place requires `drain_filter` feature.
+        self.ranges = self.ranges.iter().filter(|x| x != &dst).cloned().collect();
```
BusyJay, line 107: Why not `self.ranges.retain(|r| r != dst)`?

src/levels/compaction.rs
```diff
@@ -0,0 +1,326 @@
...
+
+        prev_ranges_len != self.ranges.len()
+    }
+
+    pub fn overlaps_with(&self, dst: &KeyRange) -> bool {
+        for r in self.ranges.iter() {
```
BusyJay, line 113: Why not `self.ranges.iter().any(|r| r.overlaps_with(dst))`?

skyzh:
I have refactored `KeyRange` as a enum, which eliminates the need of checking `is_empty` and makes the logic much clearer. Now it must be either a concrete range or infinite range. I'll take some time to check if other parts are correct after this refactor before requesting another code review, maybe this weekend :)

skyzh:
Now `KeyRange` is an enum over concrete range, `Inf` and `Empty`.

src/levels/compaction.rs
```diff
@@ -0,0 +1,425 @@
...
+use std::sync::{Arc, RwLock};
+
+use bytes::{Bytes, BytesMut};
+
+use super::LevelHandler;
+use crate::format::{key_with_ts, key_with_ts_first, key_with_ts_last, user_key};
+use crate::util::{KeyComparator, COMPARATOR};
+use crate::{Error, Result, Table};
+
+/// Represents a range of keys from `left` to `right`
```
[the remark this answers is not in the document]
skyzh: Should be `[left, right]`. Will add this to docstring.

src/levels/compaction.rs
```diff
@@ -0,0 +1,425 @@
...
+#[derive(PartialEq, Eq, Clone, Debug)]
+pub enum KeyRange {
+    Range { left: Bytes, right: Bytes },
+    Inf,
+    Empty,
```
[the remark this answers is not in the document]
skyzh: No finite range overlaps with `Empty` and all finite ranges overlap with `Inf`.

src/levels/compaction.rs
```diff
@@ -0,0 +1,425 @@
...
+                    .clone();
+                let right = if COMPARATOR.compare_key(other_right, self_right)
+                    == std::cmp::Ordering::Greater
+                {
+                    other_right
+                } else {
+                    self_right
+                }
+                .clone();
+                Range { left, right }
```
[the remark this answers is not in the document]
skyzh: Yes. When we are merging two SSTs, e.g. [2, 3] and [4, 5], the new SST spans [2, 5].

src/levels/compaction.rs
```diff
@@ -0,0 +1,425 @@
...
+            }
+        }
+    }
+
+    /// Check if two key ranges overlap
+    pub fn overlaps_with(&self, other: &Self) -> bool {
+        use KeyRange::{Empty, Inf, Range};
+        match (self, other) {
+            // If my range is empty, other ranges always overlap with me
```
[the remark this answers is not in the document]
skyzh: `overlaps_with` is not commutative, which means `a overlaps_with b` may be different from `b overlaps_with a`. And we just define the rule here to handle different cases in compaction... Generally, we would use `overlaps_with` like this: `level_keyrange.overlaps_with(&to_be_added_range)`.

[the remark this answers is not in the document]
skyzh: And the `to_be_added_range` won't be empty.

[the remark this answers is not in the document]
skyzh: In fact, I was thinking about removing `Empty` from the enum and let developers use `Option<KeyRange>`. Therefore, we could define how to handle the case when a level doesn't have any SST, each time we encounter it. Not sure which implementation is better :(

src/levels/compaction.rs
```diff
@@ -0,0 +1,425 @@
...
+
+        if !compact_def.next_range.is_empty() {
+            let next_level = &mut self.levels[next_level_id];
+            found = next_level.remove(&compact_def.next_range) && found;
+        }
+
+        if !found {
+            let this = compact_def.this_range.clone();
+            let next = compact_def.next_range.clone();
+            println!("looking for {:?} in this level", this);
```
BusyJay, line 213: Very strange output.. How about using log? Or just put it in panic.

src/levels/compaction.rs
```diff
@@ -0,0 +1,425 @@
...
+        }
+    }
+}
+
+pub fn get_key_range(tables: &[Table]) -> Option<KeyRange> {
+    if tables.is_empty() {
+        return None;
+    }
+
+    let mut smallest = tables[0].smallest().clone();
```
BusyJay, line 307: Why clone?

src/levels/compaction.rs
```diff
@@ -0,0 +1,425 @@
...
+    biggest_buf.extend_from_slice(user_key(&biggest));
+    return Some(KeyRange::new(
+        key_with_ts_first(smallest_buf),
+        // the appended key will be `<biggest_key><u64::MAX>`.
+        key_with_ts_last(biggest_buf),
+    ));
+}
+
+pub fn get_key_range_single(table: &Table) -> KeyRange {
+    let smallest = table.smallest().clone();
```
BusyJay, line 330: You can implemented as `get_key_range(slice::from_ref(table))`.

BusyJay (commented):
Rest LGTM

src/table.rs
```diff
@@ -232,8 +241,25 @@ impl TableInner {
...
+
+        let offset_length = self.offsets_length();
+        let jump = (offset_length / n).max(1);
+
+        let mut result = vec![];
+
+        for i in (0..offset_length).step_by(jump) {
+            let block = self.offsets(i).unwrap();
+            if block.key.starts_with(&prefix) {
+                result.push(Bytes::copy_from_slice(&block.key))
```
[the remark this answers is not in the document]
skyzh: Prost uses `Vec<u8>` for all byte-like types. Therefore, we always need to convert then to `Bytes` by `copy_from_slice`.

src/levels/handler.rs
```diff
@@ -0,0 +1,79 @@
+use crate::value::Value;
+use crate::Result;
+use crate::{format::user_key, get_ts, iterator_trait::AgateIterator};
```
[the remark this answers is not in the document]
skyzh: Also adds `#![allow(unused_variables)]` to eliminate warnings.

BusyJay (approved):

PR #95, opened 2021-01-06
*: lock syn version

skyzh:
`enum_dispatch` is broken with latest `syn` release `1.0.58`. To make our crate compile, we should make cargo resolver to use `syn` of `1.0.57` until `enum_dispatch` upgrades.

https://gitlab.com/antonok/enum_dispatch/-/issues/37


BusyJay (approved):
Perhaps we can remove the dependency of `enum_dispatch`.

skyzh:
A recent enum_dispatch upgrade has solved this issue.

PR #117, opened 2021-12-31
skiplist: add ut for key

zhangjinpeng1987:

**Changes**
- Add ut for "Key type in skiplist"


tabokie (approved):

skiplist/src/key.rs
```diff
@@ -50,3 +50,17 @@ impl KeyComparator for FixedLengthSuffixComparator {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_key_comparator() {
+        let suffix_comp = FixedLengthSuffixComparator::new(4);
```
BusyJay, line 60: You mean 1, right?

zhangjinpeng1987: fixed suffix "abcd", the length is 4

PR #119, opened 2022-05-10
agate: update `encode` and `decode` for `Value`

GanZiheng:

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -55,12 +55,12 @@ impl Value {
             + self.value.len()
     }
 
-    pub fn decode(&mut self, mut bytes: &mut Bytes) {
+    pub fn decode(&mut self, mut bytes: Bytes) {
         self.meta = bytes.get_u8();
         self.user_meta = bytes.get_u8();
 
         self.expires_at = decode_length_delimiter(&mut bytes).unwrap() as u64;
-        self.value = bytes.to_owned();
+        self.value = bytes;
     }
 
     pub fn encode(&self, buf: &mut BytesMut) {
@@ -159,10 +159,10 @@ mod tests {
             let mut buf = BytesMut::new();
             v.encode(&mut buf);
 
-            let mut bytes = buf.freeze();
+            let bytes = buf.freeze();
 
             let mut new_v = Value::default();
-            new_v.decode(&mut bytes);
+            new_v.decode(bytes);
 
             assert_eq!(*v, new_v);
         };
--- a/src/value.rs
+++ b/src/value.rs
@@ -211,3 +211,21 @@ impl EntryReader {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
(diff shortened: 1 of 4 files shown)


src/value.rs
```diff
@@ -211,3 +211,21 @@ impl EntryReader {
         })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_encode_decode_var() {
```
Connor1996, line 220: add check for `encode` and `decode` as well

src/value.rs
```diff
@@ -117,21 +117,19 @@ impl Value {
...
+        self.expires_at = decode_length_delimiter(&mut bytes).unwrap() as u64;
+        self.value = bytes.to_owned();
     }
 
     pub fn encode(&self, buf: &mut BytesMut) {
-        let arr: [MaybeUninit<u8>; 12] = unsafe { MaybeUninit::uninit().assume_init() };
-        let mut arr = unsafe { std::mem::transmute::<_, [u8; 12]>(arr) };
-        arr[0] = self.meta;
-        arr[1] = self.user_meta;
-        let written = encode_var(&mut arr[2..], self.expires_at);
```
Connor1996, line 133: seems we can delete `encode_var` and `decode_var` now

src/value.rs
```diff
@@ -212,3 +148,23 @@ impl EntryReader {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_encode_decode() {
+        let v = Value::new(Bytes::from("hello world"));
```
tabokie, line 158: Test `new_with_meta`

tabokie (approved):

src/value.rs
```diff
@@ -109,29 +49,25 @@ impl Value {
...
+            + self.value.len()
     }
 
-    pub fn decode(&mut self, bytes: &Bytes) {
-        self.meta = bytes[0];
-        self.user_meta = bytes[1];
-        let res = decode_var(&bytes[2..]);
-        self.expires_at = res.0;
-        self.value = bytes.slice(res.1 + 2..);
+    pub fn decode(&mut self, mut bytes: &mut Bytes) {
```
Connor1996, line 58:
```suggestion
    pub fn decode(&mut self, mut bytes: Bytes) {
```

src/value.rs
```diff
@@ -109,29 +49,25 @@ impl Value {
...
+        self.meta = bytes.get_u8();
+        self.user_meta = bytes.get_u8();
+
+        self.expires_at = decode_length_delimiter(&mut bytes).unwrap() as u64;
+        self.value = bytes.to_owned();
```
Connor1996, line 63:
```suggestion
        self.value = bytes;
```

Connor1996 (approved):
LGTM

PR #120, opened 2022-05-11
vlog: add parentheses

GanZiheng:

According to https://github.com/dgraph-io/badger/blob/master/value.go#L926 and [Expression precedence](https://doc.rust-lang.org/reference/expressions.html#expression-precedence), parentheses should be added.


GanZiheng:
/cc @Connor1996
PTAL.

skyzh (approved):
LGTM

Connor1996:
Please add test to cover this

PR #124, opened 2022-05-17
*: format imports and dependencies

Connor1996:
This PR does:
- format the imports in order by `rustfmt`
- format the dependencies in order by `cargo sort`


rust-toolchain
```diff
@@ -0,0 +1 @@
+nightly-2022-02-14
```
BusyJay, line 1: This is a bad idea. A library should not bind to a specific version.

rustfmt.toml
```diff
@@ -0,0 +1,9 @@
+version = "Two"
```
BusyJay, line 1: What's wrong with the default format?

rust-toolchain
Connor1996: ok, remove it

rustfmt.toml
Connor1996: just be consistent with tikv, not sure why change to version two before

BusyJay: I suggest to just use all the default settings of rustfmt. No one is going to sync the settings across repos when there is further change.

Connor1996: no need to sync the changes, it's just a initial to make the imports in order. We can change it further on demand regardless of how TiKV does.  Seems version two is a good practice, see https://github.com/rust-lang/rust/blob/master/rustfmt.toml#L2

BusyJay: > ..regardless of how TiKV does..

I thought you wanted "be consistent with tikv".

What do you mean "make the imports in order"?

Connor1996: `use xxx` is sorted in order

PR #123, opened 2022-05-16
Add manifest

GanZiheng:
Based on
- https://github.com/tikv/agatedb/pull/60
- https://github.com/dgraph-io/badger/blob/45bca18f24ef5cc04701a1e17448ddfce9372da0/manifest.go.

Signed-off-by: GanZiheng [ganziheng98@gmail.com](mailto:ganziheng98@gmail.com)

PR #129, opened 2022-05-24
levels: finish levels

GanZiheng:
Based on
- `levels` in https://github.com/tikv/agatedb/issues/125
- https://github.com/dgraph-io/badger/blob/45bca18f24ef5cc04701a1e17448ddfce9372da0/levels.go


GanZiheng:
Some of commits belong to https://github.com/tikv/agatedb/pull/123.

src/db/opt.rs
```diff
@@ -1,57 +1,99 @@
...
+    /*  Usually modified options. */
+    /// When set to true, the DB would call an additional sync after writes to flush
+    /// mmap buffer over to disk to survive hard reboots.
     pub sync_writes: bool,
+    /// Sets how many versions to keep per key at most.
+    pub num_versions_to_keep: usize,
+    /// When set to true, the DB will be opened on read-only mode.
+    pub read_only: bool,
+    /// When set to true, everything is stored in memory. No value/sst files are
+    /// created. In case of a crash all data will be lost.
```
zhangjinpeng1987, line 21: Please also provide the default value in the comment.

src/db/opt.rs
```diff
@@ -1,57 +1,99 @@
...
-            // memtable options
+
+            sync_writes: false,
+            num_versions_to_keep: 1,
+            read_only: false,
+            in_memory: false,
+
             mem_table_size: 64 << 20,
             base_table_size: 2 << 20,
             base_level_size: 10 << 20,
```
zhangjinpeng1987, line 79: The default value of base level size is not match with mem-table size, after a mem-table flushed to the level0, it is larger the base level size much more.

src/db/opt.rs
```diff
@@ -1,57 +1,99 @@
...
+    /// stall until compaction succeeds.
     pub num_level_zero_tables_stall: usize,
 
+    /// Sets the maximum size of a single value log file.
     pub value_log_file_size: u64,
+    /// Sets the maximum number of entries a value log file can hold approximately.
     pub value_log_max_entries: u32,
+
+    /// Sets the number of compaction workers to run concurrently.
+    pub num_compactors: usize,
```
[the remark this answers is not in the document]
GanZiheng: > Does it include the flush threads?

IMO, it doesn't include the flush threads.
According to https://github.com/dgraph-io/badger/blob/45bca18f24ef5cc04701a1e17448ddfce9372da0/db.go#L972, memtables will be flushed serially.

src/db/opt.rs
GanZiheng: >  https://github.com/GanZiheng/agatedb/blob/finish-levels/src/levels.rs#L149

For level 0, we use `mem_table_size` to determine whether it's time to do compact. And `base_level_size` is used to get the base level.

src/db/tests.rs
```diff
@@ -0,0 +1,109 @@
...
+        let mts = core.mts.read().unwrap();
+        assert_eq!(mts.nums_of_memtable(), 1);
+
+        let mt = mts.table_mut();
+
+        let key = key_with_ts(BytesMut::new(), 1);
+        let value = Value::new(Bytes::new());
+        // write_at in wal += 13
+        mt.put(key.clone(), value.clone()).unwrap();
+        mt.put(key, value).unwrap();
```
zhangjinpeng1987, line 74: Why put twice?

GanZiheng: I want to let size of memtable after `put` exceeds `value_log_file_size`, which in this test case is set to 25. I will add comment for this.

src/closer.rs
```diff
@@ -0,0 +1,58 @@
+use std::sync::{Arc, Mutex};
+
+use crossbeam_channel::{unbounded, Receiver, Sender};
+
+type CloserChanType = Arc<(Mutex<Option<Sender<()>>>, Receiver<()>)>;
```
tonyxuqqi, line 5: minor: does CloserChanType need Type suffix?

src/closer.rs
```diff
@@ -0,0 +1,58 @@
...
+        Self {
+            chan: Arc::new((Mutex::new(Some(tx)), rx)),
+        }
+    }
+
+    pub fn close(&self) {
+        self.chan.0.lock().unwrap().take();
+    }
+
+    pub fn has_been_closed(&self) -> &Receiver<()> {
```
tonyxuqqi, line 25: The name "has_been_closed" typically returns bool, to return a Receiver, why not call get_receiver?

src/levels/handler.rs
```diff
@@ -35,39 +46,187 @@ impl LevelHandler {
...
+    pub fn delete_tables(&mut self, to_del: &[Table]) -> Result<()> {
+        let mut to_del_map = HashSet::new();
+
+        for table in to_del {
+            to_del_map.insert(table.id());
+        }
+
+        let mut new_tables = vec![];
+
+        for table in &self.tables {
```
tonyxuqqi, line 77: does it need reference & here?

src/levels/handler.rs
```diff
@@ -35,39 +46,187 @@ impl LevelHandler {
...
+            if !to_del_map.contains(&table.id()) {
+                new_tables.push(table.clone());
```
tonyxuqqi, line 79: should use table instead of table.clone() if the for above does not use reference

src/levels/handler.rs
```diff
@@ -35,39 +46,187 @@ impl LevelHandler {
...
+            }
+        }
+
+        Ok(max_vs)
     }
 
-    pub fn init_tables(&mut self, tables: Vec<Table>) {
-        unimplemented!()
+    /// Appends iterators to an array of iterators, for merging.
+    pub(crate) fn append_iterators(&self, iters: &mut Vec<TableIterators>, opts: &IteratorOptions) {
```
tonyxuqqi, line 193: minor: rename it to merge_iterators?

src/levels/handler.rs
GanZiheng: Yes. Since we use `&mut self` as first parameter of this method, self.tables` is behind a mutable reference, so we cannot take ownership of it.

src/levels/handler.rs
GanZiheng: IMO, maybe it's not necessary.

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+            targets.target_size[i] = ltarget;
+            if targets.base_level == 0 && ltarget <= self.opts.base_level_size {
+                targets.base_level = i;
+            }
+            db_size /= self.opts.level_size_multiplier as u64;
+        }
+
+        let mut tsz = self.opts.base_level_size;
+
+        // Use mem_table_size for Level 0. Because at Level 0, we stop
```
zhangjinpeng1987, line 149: It's better to move this comment just above L156

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+        // the level. So, having a 1:1 size ratio between memtable size
+        // and the size of L0 files is better than churning out 32 files
+        // per memtable (assuming 64MB mem_table_size and 2MB base_table_size).
+        for i in 0..self.levels.len() {
+            if i == 0 {
+                targets.file_size[i] = self.opts.mem_table_size;
+            } else if i <= targets.base_level {
+                targets.file_size[i] = tsz;
+            } else {
+                tsz *= self.opts.table_size_multiplier as u64;
```
zhangjinpeng1987, line 160: Why not multiple with level_size_multiplier here?

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+            let del_size = {
+                let cpt_status = self.cpt_status.read().unwrap();
+                cpt_status.levels[i].del_size
+            };
+            let level = self.levels[i].read().unwrap();
+            // There could be inconsistency data which causes `size < 0`.
+            // We may safely ignore this situation.
+            // TODO: Check if we could make it more stable.
+            let size = if del_size <= level.total_size {
+                level.total_size - del_size
```
zhangjinpeng1987, line 207: Seems this policy only consider the data size doesn't deleted. How about if there are many data deleted in this level that much larger that the target size and if we don't compact this level, it will slow down the read performance, also it will cause the physical size of this level much larger than the target size of this level.

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+                0
+            };
+            add_priority(i, size as f64 / targets.target_size[i] as f64);
+        }
+
+        assert_eq!(prios.len(), self.levels.len());
+
+        // TODO: Adjust score.
+
+        prios.pop(); // remove last level
```
zhangjinpeng1987, line 218: Why not iterate to self.levels.len()-1 in L196?

zhangjinpeng1987: Tolerant for the case of self.levels is empty?

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+        let mut x: Vec<CompactionPriority> = prios.into_iter().filter(|x| x.score > 1.0).collect();
+        x.sort_by(|x, y| x.adjusted.partial_cmp(&y.adjusted).unwrap());
+        x.reverse();
```
zhangjinpeng1987, line 221: Don't need reverse if change the line above to `|x, y| y.adjusted.partial_cmp(&x.adjusted).unwrap()`

src/levels.rs
GanZiheng: Because here we calculate `file_size`, which indicates the size of SSTable, rather than the size of each level.

src/levels.rs
GanZiheng: In fact, `del_size` refers to sum of `this_size` in `CompactDef` which is being handled right now. If there are other tables being compacted, we should not reconsider it when we pick compact levels.

src/levels.rs
GanZiheng: Yes, that's a better method.

src/levels.rs
GanZiheng: > Tolerant for the case of self.levels is empty?

The number of `levels` elements will be set to `max_levels` and will never be modified. So maybe we don't have to check if it is empty.

GanZiheng: > Why not iterate to self.levels.len()-1 in L196?

That's right. We indeed will not perform compaction on bottom level.

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+                        if !right.is_empty()
+                            && COMPARATOR.compare_key(&iter_key, right) != std::cmp::Ordering::Less
+                        {
+                            break;
+                        }
+                    }
+
+                    // Ensure that all versions of the key are stored in the same
+                    // sstable, and not divided across multiple tables at the same
+                    // level.
```
zhangjinpeng1987, line 339: Seems there is no code to guarantee the constraint mentioned in the comment.

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+                    let is_expired = is_deleted_or_expired(vs.meta, vs.expires_at);
+
+                    if is_expired || last_valid_version {
+                        skip_key = BytesMut::from(&iter_key[..]);
+                        #[allow(clippy::if_same_then_else)]
+                        if !is_expired && last_valid_version {
+                            // do nothing
+                        } else if has_overlap {
+                            // do nothing
+                        } else {
```
zhangjinpeng1987, line 386: How about this key has overlap with levels below next level?

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+                ));
+            }
+
+            // TODO: Use ConcatIterator.
+            for table in valid {
+                iters.push(TableIterators::from(
+                    table.new_iterator(crate::table::ITERATOR_NOCACHE),
+                ));
+            }
+            iters
```
zhangjinpeng1987, line 477: Confused by this `iters` it contains tables in bot & top, but when construct the MergeIterator in merge_iterator.rs L184 how to guarantee `is_left_small` is true since bot tables may overlap with top tables.

GanZiheng: We can refer to `fill_tables*` to know how to fill tables for `CompactDef`. The key function inside it is `LevelHandler::overlapping_tables`, that guarantees that tables of `bot` in `CompactDef` will never be smaller than those of `top`.

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+        // Since next_level and this_level are same L0, do not need to acquire lock of next level.
+        // TODO: Don't lock cpt_status through this function.
+        let mut cpt_status = self.cpt_status.write().unwrap();
+
+        let mut out = vec![];
+
+        for table in this_level.tables.iter() {
+            if table.size() >= 2 * compact_def.targets.file_size[0] {
+                // file already big, don't include it
+                continue;
```
zhangjinpeng1987, line 600: If there are 5 sst files in L0 s1(1MB) s2(1MB) s3(64MB) s4(1MB) s5(1MB), is it will merge s1,s2,s4,s5? If so, it will break the constraint that level0's file is organized by time.

GanZiheng: We can check line 807 to see how to replace tables. And I think the problem you mentioned will not happen.

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+                if kr.overlaps_with(&dkr) {
+                    out.push(table.clone());
+                    kr = kr.extend(&dkr);
+                } else {
+                    break;
+                }
+            }
+        }
+        compact_def.this_range = get_key_range(&compact_def.top);
+        compact_def.top = out;
```
zhangjinpeng1987, line 670: Why not place L670 before L669 since L672 using the `compact_def.this_range` to calculate the overlap files?

src/levels.rs
```diff
@@ -1,5 +1,1053 @@
...
+
+            if cpt_status.overlaps_with(compact_def.next_level_id, &compact_def.next_range) {
+                continue;
+            }
+
+            if cpt_status.compare_and_add(compact_def).is_err() {
+                continue;
+            }
+
+            return Ok(());
```
zhangjinpeng1987, line 759: Seems it will only fill one compaction job at a time? Is it expected?

GanZiheng: Yes, for one specified compactor, it will compact for every level serially.

src/levels.rs
GanZiheng: That's a mistake, I will fix it, thanks.

GanZiheng: Seems like because of `Empty` range overlapping with any range, we could pass all tests. I will try to add test for this.

src/levels.rs
GanZiheng: > > Why not iterate to self.levels.len()-1 in L196?
> 
> That's right. We indeed will not perform compaction on bottom level.

Logic in adjust score will use the last level.

src/db/tests.rs
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    core.ensure_room_for_write().unwrap();
+
+    let mts = core.mts.read().unwrap();
+    assert_eq!(mts.nums_of_memtable(), 2);
+}
+
+pub fn generate_test_agate_options() -> AgateOptions {
```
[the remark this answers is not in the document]
GanZiheng: `generate_test_agate_options` just generate the option.

src/levels.rs
GanZiheng: > How about this key has overlap with levels below next level?

Sorry I misunderstood what you were asking. As for the current key, it will be added to the table, although it is deleted or expired.

src/opt.rs
```diff
@@ -19,3 +24,20 @@ pub enum ChecksumVerificationMode {
...
+}
+
+pub fn build_table_options(opt: &crate::AgateOptions) -> TableOptions {
+    // get latest data key
+    TableOptions {
+        table_size: opt.base_table_size,
+        block_size: opt.block_size,
+        bloom_false_positive: opt.bloom_false_positive,
+        checksum_mode: opt.checksum_mode,
+        table_capacity: (opt.base_level_size as f64 * 0.95) as u64,
```
zhangjinpeng1987, line 41: Why not `table_size` but `base_level_size`?

src/table/builder.rs
```diff
@@ -183,7 +183,7 @@ impl Builder {
         let cs = self.build_checksum(&bytes);
         self.write_checksum(cs);
         // TODO: eliminate clone if we do not need builder any more after finish
-        self.buf.clone().freeze()
+        self.buf.freeze()
```
zhangjinpeng1987, line 186: Please also remove comment in L185.

src/util.rs
```diff
@@ -71,3 +78,15 @@ pub fn same_key(a: &[u8], b: &[u8]) -> bool {
     }
     return user_key(a) == user_key(b);
 }
+
+pub fn unix_time() -> u64 {
+    let start = SystemTime::now();
```
zhangjinpeng1987, line 83: Can we use `coarsetime` to improve the performance?

src/opt.rs
GanZiheng: Sorry, but what `table_size`? I think there is no problem we set `table_size` to `base_table_size` here. We will either change or not use this field later.

src/table/builder.rs
GanZiheng: Ok.

zhangjinpeng1987 (approved):
LGTM

src/db/opt.rs
```diff
@@ -1,59 +1,143 @@
...
+    /// If it doesn't exist, Agate will try to create it for you.
     pub value_dir: PathBuf,
-    // TODO: docs
-    pub in_memory: bool,
+
+    /*  Usually modified options. */
+    /// When set to true, the DB would call an additional sync after writes to flush
+    /// mmap buffer over to disk to survive hard reboots.
+    ///
+    /// The default value of `sync_writes` is false.
```
Connor1996, line 20: The default value is defined below, seem no need to write again here

src/ops.rs
```diff
@@ -1,3 +1,3 @@
-mod oracle;
+pub mod oracle;
```
Connor1996, line 1: define these in `mod.rs` in ops dir instead of this file

src/levels/compaction.rs
```diff
@@ -298,9 +307,9 @@ impl Targets {
     }
 }
 
-pub fn get_key_range(tables: &[Table]) -> Option<KeyRange> {
+pub fn get_key_range(tables: &[Table]) -> KeyRange {
     if tables.is_empty() {
-        return None;
+        return KeyRange::default();
```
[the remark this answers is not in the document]
GanZiheng: One member in `KeyRange` is `Empty`.

src/db/opt.rs
GanZiheng: There're many fields, so it's not convenient to look for default value for a certain field. Besides, if we add comment, we could easily get the value via language server such as rust-analyzer.

src/ops.rs
GanZiheng: I think that's a better way, I will update the code.

src/levels/handler.rs
```diff
@@ -35,39 +46,187 @@ impl LevelHandler {
...
-    }
+            let mut it = table.new_iterator(0);
+            it.seek(key);
+            if !it.valid() {
+                continue;
+            }
 
-    pub fn delete_tables(&mut self, to_del: &[Table]) -> Result<()> {
-        unimplemented!()
+            if crate::util::same_key(key, it.key()) {
```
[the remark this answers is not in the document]
skyzh: This depends on GC policy. In AgateDB, we need to search all levels.

src/error.rs
```diff
@@ -88,4 +90,11 @@ impl<T: Sized> From<PoisonError<T>> for Error {
     }
 }
 
+impl From<Box<dyn Any + Send>> for Error {
```
skyzh, line 93: I don'e think we should be able to convert arbitrary `Box<dyn Any + Send>` to Error. We should explicitly write `.map_err(Error::JoinError)` when necessary.

src/levels.rs
```diff
@@ -1,5 +1,1068 @@
...
+            // Update the manifest _before_ the table becomes part of a level_handler.
+            self.manifest
+                .add_changes(vec![new_create_change(table.id(), 0, 0)])?;
+        }
+
+        while !self.levels[0].write()?.try_add_l0_table(table.clone()) {
+            let time_start = std::time::Instant::now();
+
+            while self.levels[0].read()?.num_tables() >= self.opts.num_level_zero_tables_stall {
+                std::thread::sleep(std::time::Duration::from_millis(10));
```
skyzh, line 904: In the future, we can use something like notifier or condition variable to avoid busy waiting loop.

skyzh (approved):
Rest LGTM from my side, please resolve others' comments.

src/error.rs
```diff
@@ -51,6 +51,8 @@ pub enum Error {
     ReadOnlyError(String),
     #[error("Lock Poison")]
     PoisonError(String),
+    #[error("Join Error")]
+    JoinError(Box<dyn Any + Send>),
```
Connor1996, line 55: where is it used

src/levels/compaction.rs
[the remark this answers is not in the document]
Connor1996: I mean why not remove the `Empty` item in `KeyRange`

src/ops/oracle.rs
```diff
@@ -21,4 +22,9 @@ impl Oracle {
     pub fn set_discard_ts(&self, discard_ts: u64) {
         self.discard_ts.store(discard_ts, Ordering::SeqCst);
     }
+
+    pub(crate) fn discard_at_or_below(&self) -> u64 {
+        // TODO: Implement it.
```
Connor1996, line 27: unimplemented!()

src/levels.rs
```diff
@@ -1,5 +1,1068 @@
...
+    opts: AgateOptions,
+    orc: Arc<Oracle>,
+
+    cpt_status: RwLock<CompactStatus>,
+    manifest: Arc<ManifestFile>,
+    // TODO: Add l0_stalls_ms.
+}
+
+pub struct LevelsController {
+    pub(crate) core: Arc<Core>,
```
Connor1996, line 52: How about `LevelsControllerInner`?

src/levels.rs
```diff
@@ -1,5 +1,1068 @@
...
+
+        let manifest_data = manifest.manifest_cloned();
+
+        let mut max_file_id = 0;
+        let mut tables: Vec<Vec<Table>> = Vec::new();
+        tables.resize(opts.max_levels, vec![]);
+
+        // TODO: Parallelly open tables.
+        for (id, table_manifest) in manifest_data.tables {
+            if id > max_file_id {
```
Connor1996, line 93: use `std::cmp::max`

src/levels.rs
```diff
@@ -1,5 +1,1068 @@
...
+        if opts.in_memory {
+            return Ok(core);
+        }
+
+        // TODO: Revert to manifest.
+
+        let manifest_data = manifest.manifest_cloned();
+
+        let mut max_file_id = 0;
+        let mut tables: Vec<Vec<Table>> = Vec::new();
```
Connor1996, line 88: vec![vec![]; opts.max_levels]

src/levels.rs
```diff
@@ -1,5 +1,1068 @@
...
+            };
+
+            tables.push(table);
+        }
+
+        info!(
+            "compactor {}, sub_compact took {} mills, produce {} tables, added {} keys, skipped {} keys",
+            compact_def.compactor_id,
+            std::time::Instant::now()
+                .duration_since(start_time)
```
Connor1996, line 444: start_time.elapsed()

src/levels.rs
```diff
@@ -1,5 +1,1068 @@
...
+        let has_overlap =
+            self.check_overlap(&compact_def.all_tables(), compact_def.next_level_id + 1);
+
+        let discard_ts = self.orc.discard_at_or_below();
+
+        // TODO: Collect stats for GC.
+
+        // TODO: Check if the given key-range overlaps with next_level too much.
+
+        let start_time = std::time::Instant::now();
```
Connor1996, line 261: coarse time should be enough

src/ops.rs
GanZiheng: According to the note of https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames, we should stay unchanged.

src/levels/compaction.rs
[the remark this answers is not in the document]
GanZiheng: I think it is a better way to make `Empty` as one kind of `KeyRange`. Otherwise, we should use `Option<KeyRange>` as parameter instead, in which case we should check if it is `None` first and then do match with `KeyRange`, I just think it is unnecessary.

src/error.rs
GanZiheng: In `compact_build_tables`.

src/ops/oracle.rs
GanZiheng: Current we need to use it.

src/levels.rs
GanZiheng: I added comment for it.

src/levels.rs
GanZiheng: Maybe unnecessary.

Connor1996 (approved):
LGTM

PR #130, opened 2022-05-31
compaction: define the variable just before it used

zhangjinpeng1987:

define the variable just before it used


Connor1996 (approved):
LGTM

PR #131, opened 2022-05-31
ci: Introduce code coverage check

Connor1996:
Use [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov#basic-usage) to generate LLVM source-based code coverage.


skyzh (approved):

skyzh:
Seems that the codecov page isn't made public?

Connor1996:
> Also no coverage is uploaded 🤣 need to set codecov token?

It says no coverage report for the base to make the difference, but not this PR. I believe it will have a result in the next PR. Codecov token is used for private repo as I understand.

BusyJay (approved):

PR #135, opened 2022-06-02
vlog: hold lock as short as we can

zhangjinpeng1987:


skyzh (approved):
Actually, this function is only used on startup when init the database. But I'd okay with any optimization to make this process a little bit faster.

PR #136, opened 2022-06-02
txn: finish oracle

GanZiheng:
Based on
- https://github.com/tikv/agatedb/pull/67
- https://github.com/dgraph-io/badger/blob/45bca18f24ef5cc04701a1e17448ddfce9372da0/txn.go


src/ops/oracle.rs
```diff
@@ -1,30 +1,312 @@
...
+                .any(|committed_txn| {
+                    reads
+                        .iter()
+                        .any(|read| committed_txn.conflict_keys.contains(read))
+                })
+        }
+    }
+}
+
 pub struct Oracle {
```
Connor1996, line 72: What's the oracle used for?

GanZiheng: It is used for transaction to get timestamp when reading or committing.

Connor1996: add a comment

src/levels/tests.rs
```diff
@@ -31,6 +31,9 @@ where
     F: FnOnce(&mut LevelsController),
     F: Send + 'static,
 {
+    // So as to set and get discard timestamp by self.
+    opts.managed_txns = true;
```
Connor1996, line 35: set it in `test_options()`?

src/ops/oracle.rs
```diff
@@ -1,30 +1,312 @@
...
+            oracle.done_commit(commit_ts);
+        }
+
+        oracle.txn_mark.wait_for_mark(100);
+
+        oracle.stop();
+    }
+
+    #[test]
+    fn test_detect_confict() {
```
Connor1996, line 266: confict -> conflict

src/ops/oracle.rs
```diff
@@ -1,30 +1,312 @@
...
+    discard_ts: u64,
+    /// Contains all committed writes (contains fingerprints of keys written
+    /// and their latest commit counter).
+    committed_txns: Vec<CommittedTxn>,
+    last_cleanup_ts: u64,
+}
+
+impl CommitInfo {
+    fn cleanup_committed_transactions(&mut self, is_managed: bool, read_mark: Arc<WaterMark>) {
+        let max_read_ts = {
```
Connor1996, line 30: unnecessary bracket

skyzh:
I think we can decide whether to have the "managed mode" in agatedb. I believe systems built upon agatedb generally will have their own timestamp and way of watermark management. We can simply design agatedb as a user-timestamp system, and only have managed mode. This way we don't need conflict detector.

GanZiheng:
> I think we can decide whether to have the "managed mode" in agatedb. I believe systems built upon agatedb generally will have their own timestamp and way of watermark management. We can simply design agatedb as a user-timestamp system, and only have non-managed mode. This way we don't need conflict detector.

If I understand correctly, systems with their own timestamp generator could just set timestamp in managed mode?

Connor1996:
> I think we can decide whether to have the "managed mode" in agatedb. I believe systems built upon agatedb generally will have their own timestamp and way of watermark management. We can simply design agatedb as a user-timestamp system, and only have non-managed mode. This way we don't need conflict detector.

We can first port everything without tweaks, then truncate the unnecessary part.

zhangjinpeng1987 (approved):
LGTM

Connor1996 (approved):
LGTM

PR #138, opened 2022-06-07
flush memtable to L0

zhangjinpeng1987:

* Part of https://github.com/tikv/agatedb/issues/125, pick https://github.com/tikv/agatedb/pull/47 to master
* Fix the bug of levelHander::get data from sst & db.get data from memtable will get nothing when the value.version = 0

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -227,6 +227,7 @@ impl Core {
                 if value.version == version {
                     return Ok(value);
                 }
+                // max_vs.version == 0 means it is not assigned a value yet.
                 if max_value.version == 0 || max_value.version < value.version {
                     max_value = value;
                 }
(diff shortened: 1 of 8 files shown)


src/db/opt.rs
```diff
@@ -105,6 +105,10 @@ pub struct AgateOptions {
     ///
     /// The default value of `checksum_mode` is [`ChecksumVerificationMode`].
     pub checksum_mode: opt::ChecksumVerificationMode,
+    /// Create the directory if the provided open path doesn't exists.
```
Connor1996, line 108: add a blank

src/db/opt.rs
```diff
@@ -161,4 +166,55 @@ impl AgateOptions {
         // TODO: take other options into account
         self.mem_table_size as u64
     }
+
+    pub fn create(&mut self) -> &mut AgateOptions {
+        self.create_if_not_exists = true;
+        self
+    }
+
+    pub fn num_memtables(&mut self, num_memtables: usize) -> &mut AgateOptions {
```
Connor1996, line 175: how about using https://github.com/Hoverbear/getset

src/memtable.rs
```diff
@@ -194,18 +200,26 @@ impl MemTables {
     }
 
     /// Get mutable memtable
-    pub fn table_mut(&self) -> &MemTable {
-        &self.mutable
+    pub fn table_mut(&self) -> Arc<MemTable> {
```
Connor1996, line 203: `xxx_mut` usually means a mutable refernece of xxx, better rename to `mut_table`

src/db.rs
```diff
@@ -140,51 +242,225 @@ impl Core {
...
+                Ok(())
+            }
+            Err(_) => Err(Error::WriteNoRoom(())),
+        }
+    }
 
-        let memtable = self.new_mem_table()?;
+    /// build L0 table from memtable
+    fn build_l0_table(ft: FlushTask, table_opts: TableOptions) -> TableBuilder {
+        let mut iter = ft.mt.skl.iter_ref();
```
Connor1996, line 328: don't think these namings have a good readiability

src/db/opt.rs
zhangjinpeng1987: make format said no

src/db/opt.rs
zhangjinpeng1987: In the early day I don't want to introduce too many dependencies.

src/memtable.rs
```diff
@@ -194,18 +200,26 @@ impl MemTables {
...
+        self.mutable.clone()
+    }
+
+    pub fn table_imm(&self, idx: usize) -> Arc<MemTable> {
```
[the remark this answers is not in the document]
zhangjinpeng1987: done

src/db/opt.rs
zhangjinpeng1987: done

src/db/opt.rs
Connor1996: there are lots of configs, it's verbose to add setter for each of them mannually

zhangjinpeng1987: How about another dedicated PR to do it?

Connor1996: okay

src/db.rs
```diff
@@ -140,51 +242,227 @@ impl Core {
...
+        }
+
+        Ok(())
+    }
+
+    /// Write requests should be only called in one thread. By calling this
+    /// function, requests will be written into the LSM tree.
+    ///
+    /// TODO: ensure only one thread calls this function by using Mutex.
+    #[allow(clippy::needless_collect)]
```
Connor1996, line 406: can we avoid it?

src/db.rs
```diff
@@ -140,51 +242,227 @@ impl Core {
...
+        };
 
-        debug!(
-            "memtable flushed, total={}, mt.size={}",
-            mts.nums_of_memtable(),
-            mem_size
-        );
+        self.lvctl.add_l0_table(table)?;
 
         Ok(())
```
Connor1996, line 376: no need, return `self.lvctl.add_l0_table(table)` directly

Connor1996 (commented):
rest LGTM

src/db.rs
zhangjinpeng1987: No, the clippy will say:
```
error: avoid using `collect()` when not needed
   --> src/db.rs:409:69
    |
409 |         let dones: Vec<_> = requests.iter().map(|x| x.done.clone()).collect();
    |                                                                     ^^^^^^^
...
442 |         for done in dones.into_iter().flatten() {
    |                     ----------------- the iterator could be used here instead
```

Connor1996 (approved):
LGTM

GanZiheng (approved):
LGTM

PR #140, opened 2022-06-08
ci: use stable rust version

zhangjinpeng1987:


BusyJay (approved):

.github/workflows/test.yml
```diff
@@ -76,7 +76,7 @@ jobs:
         name: Setup Cargo Toolchain 🛎️
         with:
           components: rustfmt, clippy, rust-src
-          toolchain: nightly
+          toolchain: stable
```
BusyJay, line 79: Running sanitize test should use nightly toolchain

zhangjinpeng1987: got it

PR #142, opened 2022-06-08
test: fix table checksum test

GanZiheng:
`table::tests::test_table_checksum` truncates data of SST randomly and asserts if we read the truncated data, we will get `InvalidChecksum` error. But if checksum length part of data truncated, we may get wrong length value even larger than whole data length while parsing the data, and that will result in subtraction overflow panic.


Connor1996 (commented):
Please paste the possible failure in PR description

zhangjinpeng1987 (approved):
LGTM

PR #143, opened 2022-06-09
txn: finish transaction

GanZiheng:


GanZiheng:
A quite weird problem on leak sanitizer test.
As shown in https://github.com/tikv/agatedb/runs/6933910360?check_suite_focus=true, segmentation violation occurred while tests were running. And if we test it locally, we could additionally get error information as below.
screenshot-20220617-175736
Currently I have no idea about why this error occurs. It is not reproduced stably. I have tried a lot to find which modification introduces this error, and I found it may have relationship with the compaction. If we set option to run compaction more frequently, we then be more easier to get this error.
And another point that attracts me is `premature terminator entry` in the picture above. I guess some codes make a memory corruption when running.

.github/workflows/test.yml
```diff
@@ -82,7 +82,7 @@ jobs:
         name: Running Tests with ${{ matrix.sanitizer }} sanitizer 🚀
         with:
           command: test
-          args: --all-features -Zbuild-std --target x86_64-unknown-linux-gnu
+          args: --all-features -Zbuild-std --target x86_64-unknown-linux-gnu -- --test-threads=1
```
zhangjinpeng1987, line 85: setting thread=1 for what?

zhangjinpeng1987 (commented):
Please fix conflicts.

.github/workflows/test.yml
GanZiheng: I found it could resolve problem in leak sanitizer test locally, but somehow doesn't work on github workflow.

src/ops/transaction_test.rs
```diff
@@ -842,31 +842,29 @@ mod normal_db {
         where
             F: FnOnce(Arc<Agate>) + Clone + Send + 'static,
         {
-            let num_go = 16;
+            let num_go = 4;
```
Connor1996, line 845: Add a comment for the encounter issue

src/db.rs
```diff
@@ -471,20 +647,6 @@ impl Agate {
     pub fn write_requests(&self, request: Vec<Request>) -> Result<()> {
```
Connor1996, line 647: move these to the top impl block?

src/ops/transaction.rs
```diff
@@ -1,51 +1,559 @@
...
+    /// Used in managed mode to store duplicate entries.
+    pub(crate) duplicate_writes: Vec<Entry>,
+
+    pub(crate) num_iterators: AtomicUsize,
+    pub(crate) discarded: bool,
     pub(crate) done_read: bool,
+    // Update is used to conditionally keep track of reads.
+    pub(crate) update: bool,
+
+    pub(crate) agate: Arc<crate::db::Core>,
```
Connor1996, line 48: rename to `core`

Connor1996 (commented):
rest LGTM

src/db.rs
GanZiheng: I think stay there is enough :rofl:

Connor1996: why separate impl block

GanZiheng: Sorry I'm confused, what separates the impl block?

GanZiheng: I merged two impl block for `Agate`.

.github/workflows/test.yml
GanZiheng: I set `test-threads=1` with reducing concurrency of `ops::oracle::tests::test_detect_conflict`, and passed the leak sanitizer test.

Connor1996 (approved):
LGTM

PR #145, opened 2022-06-14
db/opt: use setget lib instead of set/get methods

zhangjinpeng1987:


Connor1996 (approved):
LGTM

PR #148, opened 2022-06-14
db: start bg compaction threads after init agate

zhangjinpeng1987:


GanZiheng (approved):
LGTM

PR #151, opened 2022-06-14
db/opt: adjust arena size based on mem-table-size to prevent arena grow

zhangjinpeng1987:
adjust arena size based on mem-table-size to prevent arena grow in most cases.


Connor1996 (approved):
LGTM

GanZiheng (approved):
LGTM

PR #158, opened 2022-06-22
skiplist: not use CAS for single thread writing

zhangjinpeng1987:


skyzh (commented):
In what cases will we concurrently write to memtable?

https://github.com/tikv/agatedb/blob/366404d0a4960eeddd8e5dfd1af318716e621cfd/src/db.rs#L248-L254

Currently, all writes go into only one thread.

zhangjinpeng1987:
> In what cases will we concurrently write to memtable?
> 
> https://github.com/tikv/agatedb/blob/366404d0a4960eeddd8e5dfd1af318716e621cfd/src/db.rs#L248-L254
> 
> Currently, all writes go into only one thread.

@skyzh The context is that, skiplist's previous implementation works under both single-thread write and multi-threads write. But there is a more efficient way to implement for single thread write scenario which just use atomic.store, don't use CAS.

> Currently, all writes go into only one thread.
This is true, but skiplist is a fundamental component, maybe in the future we need the concurrent write ability.

skiplist/src/list.rs
```diff
@@ -223,7 +225,7 @@ impl<C: KeyComparator> Skiplist<C> {
         }
```
GanZiheng, line 225: Should we also change the logic of increasing height if we add `allow_concurrent_write` option?

zhangjinpeng1987: done

zhangjinpeng1987:
@skyzh Any other concerns or comments?

skyzh (approved):

GanZiheng (approved):
LGTM

skiplist/src/list.rs
```diff
@@ -210,20 +212,68 @@ impl<C: KeyComparator> Skiplist<C> {
...
+                    height,
+                    Ordering::SeqCst,
+                    Ordering::SeqCst,
+                ) {
+                    Ok(_) => break,
+                    Err(h) => list_height = h,
+                }
+            }
+        } else {
+            // There is no need to use CAS for single thread writing.
```
Connor1996, line 228: we plan to support the multiple thread writing, it's necessary now? We may need to delete in future.

zhangjinpeng1987: If treat AgateDB as a fundamental kv engine on cloud, maybe it need both single thread write and multi-threads write.

PR #163, opened 2022-06-24
skiplist: doubly list [20x faster reverse scan]

zhangjinpeng1987:

close https://github.com/tikv/agatedb/issues/134

Implement doubly link of skiplist for the single-thread write scenario. It makes the reverse scan 20x faster than before.
```
skiplist_reverse_scan   time:   [1.5644 ms]
doubly_skiplist_reverse_scan  time:   [75.162 us]
```

### DIFF
--- a/skiplist/benches/bench.rs
+++ b/skiplist/benches/bench.rs
@@ -163,10 +163,50 @@ fn bench_write_skiplist(c: &mut Criterion) {
     j.join().unwrap();
 }
 
+fn bench_reverse_scan_skiplist(c: &mut Criterion) {
+    let comp = FixedLengthSuffixComparator::new(8);
+    let list = Skiplist::with_capacity(comp, 16 << 20, true);
+    let value = Bytes::from_static(b"0123456789abcdefghijklmkopqrstuvwxyz");
+    let mut rng = rand::thread_rng();
+    for _ in 0..10000 {
+        list.put(random_key(&mut rng), value.clone());
+    }
+    c.bench_function("skiplist_reverse_scan", |b| {
+        b.iter(|| {
+            let mut iter = list.iter_ref();
+            iter.seek_to_last();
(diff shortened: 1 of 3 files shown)


skiplist/src/list.rs
```diff
@@ -431,8 +444,20 @@ impl<T: AsRef<Skiplist<C>>, C: KeyComparator> IterRef<T, C> {
 
     pub fn prev(&mut self) {
         assert!(self.valid());
-        unsafe {
-            self.cursor = self.list.as_ref().find_near(self.key(), true, false);
+        if self.list.as_ref().allow_concurrent_write {
+            unsafe {
+                self.cursor = self.list.as_ref().find_near(self.key(), true, false);
+            }
```
GanZiheng, line 450: Can we apply this improvement when concurrent write is allowed?

skiplist/benches/bench.rs
```diff
@@ -163,10 +163,50 @@ fn bench_write_skiplist(c: &mut Criterion) {
...
+            let mut iter = list.iter_ref();
+            iter.seek_to_last();
+            while iter.valid() {
+                iter.prev();
+            }
+        })
+    });
+}
+
+fn bench_reverse_scan_skiplist_fast(c: &mut Criterion) {
```
Connor1996, line 185: what's the different with non-fast

skiplist/src/list.rs
zhangjinpeng1987: It is much more complicated for concurrent write, ref to https://github.com/tikv/rocksdb/pull/278/files, maybe in the future we can port it.

Connor1996 (approved):
LGTM

skiplist/src/list.rs
GanZiheng: I see, exactly not easy :joy:

PR #168, opened 2022-06-27
chore: update badger link

Connor1996:
[Dgraph is dissolved](https://discuss.dgraph.io/t/quitting-dgraph-labs/16702) and the original badger is not maintained anymore. Now the badger is maintained under the newly found org, see https://github.com/outcaste-io/badger


BusyJay:
Let's keep the original link for now and see how things go. On the other hand, we are expected to split from the original implementation and optimize on our own for our specific cases, so the link is not important unless the repo is deleted.

zhangjinpeng1987 (approved):
LGTM

zhangjinpeng1987:
@Connor1996 How about close it according to jay's suggestion?

Connor1996:
> Let's keep the original link for now and see how things go. On the other hand, we are expected to split from the original implementation and optimize on our own for our specific cases, so the link is not important unless the repo is deleted.

@BusyJay The situation is clear, the old repo is out of maintenance due to politics, and the core authors establish a new org to maintain it. You can check the article written by the CEO.

PR #170, opened 2022-06-27
agate: support write batch

GanZiheng:


src/batch.rs
```diff
@@ -0,0 +1,185 @@
...
+        let mut new_txn = self.core.new_transaction(true, self.is_managed);
+        new_txn.commit_ts = self.commit_ts;
+        let txn = std::mem::replace(&mut self.txn, new_txn);
+        txn.commit()?;
+
+        Ok(())
+    }
+
+    fn delete(&mut self, key: Bytes) -> Result<()> {
+        if let Err(err) = self.txn.delete(key.clone()) {
```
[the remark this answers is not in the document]
GanZiheng: I think we should also set `self.err` here. I will update this.

src/batch.rs
```diff
@@ -0,0 +1,187 @@
...
+                self.err = Some(err.clone());
+                return Err(err);
+            }
+        } else {
+            return Ok(());
+        }
+
+        self.commit()?;
+
+        if let Err(err) = self.txn.set_entry(entry) {
```
Connor1996, line 81: I don't understand, why `set_entry` twice

GanZiheng: At first `set_entry`, we may get `TxnTooBig` error. In that case, we should commit and then `set_entry` again.

Connor1996 (commented):
the coverage is not adequate

src/db.rs
```diff
@@ -408,10 +409,14 @@ impl Core {
...
-            }
-            builder.add(iter.key(), &vs, 0); // TODO: support vlog length
+            let vlog_len = if vs.meta & value::VALUE_POINTER != 0 {
+                let mut vp = ValuePointer::default();
+                vp.decode(&vs.value);
+                vp.len
+            } else {
+                0
+            };
+            builder.add(iter.key(), &vs, vlog_len); // TODO: support vlog length
```
Connor1996, line 419: remove the TODO

Connor1996 (approved):
rest LGTM

PR #171, opened 2022-06-27
levels: don't sync directory in memory mode

GanZiheng:
This will cause problem with heavy workload in memory mode, since we will never return tables built by compaction successfully.

### DIFF
--- a/src/levels.rs
+++ b/src/levels.rs
@@ -524,7 +524,10 @@ impl LevelsControllerInner {
             new_tables.append(&mut table?);
         }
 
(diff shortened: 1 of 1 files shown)


Connor1996 (commented):
What's the error

GanZiheng:
> What's the error

If we put entries enough to cause compaction in memory mode, we would get this error.

I tried to add test for this, but it will result in OOM in memory sanitizer test.

Connor1996 (approved):
LGTM

PR #173, opened 2022-07-04
bench: add bench action

GanZiheng:

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -2,9 +2,15 @@ on:
   workflow_dispatch:
     inputs:
       reason:
-        description: 'reason to trigger this build'
+        description: "reason to trigger this build"
         required: false
-  
+
+permissions:
+  # deployments permission to deploy GitHub pages website
+  deployments: write
+  # contents permission to update benchmark contents in gh-pages branch
+  contents: write
+
 name: Bench
 
 jobs:
@@ -26,11 +32,17 @@ jobs:
         with:
           toolchain: nightly
           default: true
-      - uses: actions-rs/cargo@v1
-        name: Benchmark 🚀
+      - name: Benchmark 🚀
+        run: cargo +nightly bench --all-features --workspace -- --output-format bencher | tee output.txt
+      - uses: benchmark-action/github-action-benchmark@v1
+        name: Store benchmark result
         with:
-          command: bench
-          args: --all-features --workspace  
+          tool: "cargo"
+          output-file-path: output.txt
+          # Access token to deploy GitHub Pages branch
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Push and deploy GitHub pages branch automatically
+          auto-push: true
   sanitizer_bench:
     name: Bench with Sanitizer
     runs-on: ubuntu-latest
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,3 +54,6 @@ incremental = false
 debug-assertions = false
 overflow-checks = false
 rpath = false
+
+[lib]
+bench = false
--- a/proto/Cargo.toml
+++ b/proto/Cargo.toml
@@ -10,3 +10,6 @@ prost = "0.8"
 
 [build-dependencies]
 prost-build = { version = "0.8" }
+
+[lib]
+bench = false
--- a/skiplist/Cargo.toml
+++ b/skiplist/Cargo.toml
@@ -18,3 +18,6 @@ tikv-jemallocator = "0.4.0"
 [[bench]]
 name = "bench"
 harness = false
+
+[lib]
+bench = false


Connor1996 (commented):
Where is the output

.github/workflows/bench.yml
```diff
@@ -26,55 +31,14 @@ jobs:
...
-      - uses: actions-rs/cargo@v1
-        name: Bench with Thread Sanitizer 🚀
-        with:
-          command: bench
-          args: --all-features -Zbuild-std --target x86_64-unknown-linux-gnu
-        env:
-          RUSTFLAGS: "-Zsanitizer=thread"
\ No newline at end of file
+      - name: Benchmark 🚀
+        run: cargo +nightly bench --all-features --workspace -- --output-format bencher | tee output.txt
```
Connor1996, line 35: run the cases of badger bench instead of micro bench

benches/benches_badger_rocks.rs
```diff
@@ -9,7 +9,10 @@ use criterion::{criterion_group, criterion_main, Criterion};
 use rand::prelude::ThreadRng;
 use rand::Rng;
 use rocksdb::DB;
+use std::ops::Add;
```
Connor1996, line 12: use nightly toolchain, they are not formatted

Cargo.toml
```diff
@@ -60,3 +69,6 @@ incremental = false
 debug-assertions = false
 overflow-checks = false
 rpath = false
+
+[lib]
+bench = false
```
Connor1996, line 74: why

benches/benches_badger_rocks.rs
```diff
@@ -0,0 +1,255 @@
...
+        });
+    });
+
+    dir.close().unwrap();
+    let dir = TempDir::new("agatedb-bench-large-value").unwrap();
+    let dir_path = dir.path();
+    opts.dir = dir_path.to_path_buf();
+    opts.value_dir = dir_path.to_path_buf();
+
+    c.bench_function("badger populate large value", |b| {
```
Connor1996, line 141: should reuse the `bench_badger` impl with value size as a variable.

Cargo.toml
GanZiheng: Because  `--output-format` is an argument that Criterion.rs supports but libtest does not.

According to https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options, we could either disable benchmarks for crate or run only Criterion benchmark.

Since I have used the latter method due to only benching `benches_badger_rocks`, I will restore these modifications on Cargo.toml.

benches/benches_badger_rocks.rs
GanZiheng: In fact, I found we could not simply reuse them.

GanZiheng:
This is screenshot of GitHub page about benchmark result.
screenshot-20220706-160608

Connor1996:
> This is screenshot of GitHub page about benchmark result. screenshot-20220706-160608

can we merge the same case for badger and rocks in one diagram?

GanZiheng:
> 

Seems we cannot do it :rofl:

GanZiheng:
> can we merge the same case for badger and rocks in one diagram?

After modifying the html file generated by the action, we could get something like this.
screenshot-20220706-203812

benches/benches_agate_rocks.rs
```diff
@@ -0,0 +1,253 @@
...
+};
+
+use agatedb::{Agate, AgateOptions, IteratorOptions};
+use common::{gen_kv_pair, remove_files, unix_time};
+use criterion::{criterion_group, criterion_main, Criterion};
+use rand::{prelude::ThreadRng, Rng};
+use rocksdb::DB;
+use tempdir::TempDir;
+
+const BATCH_SIZE: u64 = 1000;
```
Connor1996, line 17: is this too small?

benches/benches_agate_rocks.rs
```diff
@@ -0,0 +1,253 @@
...
+const SMALL_VALUE_SIZE: usize = 32;
+const LARGE_VALUE_SIZE: usize = 102400;
```
Connor1996, line 19: 1000 is enough

benches/benches_agate_rocks.rs
```diff
@@ -0,0 +1,253 @@
...
+
+    for i in 0..BATCH_SIZE {
+        let (key, value) = gen_kv_pair(i, value_size);
+        txn.set(key, value).unwrap();
+    }
+
+    txn.commit_at(unix_time()).unwrap();
+}
+
+fn agate_randread(agate: Arc<Agate>, value_size: usize, rng: &mut ThreadRng) {
```
Connor1996, line 32: can we reuse the logic of agate bench and support multiple threads

benches/benches_agate_rocks.rs
GanZiheng: I set it to 4096.

benches/benches_agate_rocks.rs
GanZiheng: I have updated this, PTAL.

Connor1996 (approved):
LGTM

PR #184, opened 2022-07-21
iterator: uniform behavior and implement `prev` and `to_last` to all Iterators

GanZiheng:

Uniform all kinds of iterators which implement the `AgateIterator` trait.

Suppose we have another two extra elements in the data which the iterator iterates over, one is before the original first element, called `before first`, and another is after the last original element, namely `after last`.

1. If we arrives the first element, and call `prev`, we will arrive `before first`, and `valid` will return false;
2. If we arrives the last element, and call `next`, we will arrive `after last`, and `valid` will return false;
3. At `before first`, if  we call `prev`, nothing will change, and if we call `next`, we will move to the first element;
4. At `after last`, if  we call `next`, nothing will change, and if we call `prev`, we will move to the last element;

### DIFF
--- a/src/iterator.rs
+++ b/src/iterator.rs
@@ -299,6 +299,18 @@ impl<'a> Iterator<'a> {
         self.item = None;
     }
 
+    pub fn prev(&mut self) {
+        self.table_iter.seek(&self.last_key.clone().freeze());
+        self.table_iter.prev();
+        while self.table_iter.valid() {
+            if self.parse_item_for_prev() {
+                return;
+            }
+        }
+
+        self.item = None;
+    }
+
     /// Handles both forward and reverse iteration implementation. We store keys such that
     /// their versions are sorted in descending order. This makes forward iteration
     /// efficient, but revese iteration complicated. This tradeoff is better because
@@ -332,20 +344,20 @@ impl<'a> Iterator<'a> {
             return true;
         }
 
-        // If iterating in forward direction, then just checking the last key against
-        // current key would be sufficient.
-        if !self.opt.reverse {
-            if crate::util::same_key(&self.last_key, key) {
-                self.table_iter.next();
-                return false;
-            }
-            // Only track in forward direction.
-            // We should update last_key as soon as we find a different key in our snapshot.
-            // Consider keys: a 5, b 7 (del), b 5. When iterating, last_key = a.
-            // Then we see b 7, which is deleted. If we don't store last_key = b, we'll then
-            // return b 5, which is wrong. Therefore, update last_key here.
+        // To support prev, we also should track last_key in forward direction.
+        let key_changed = if !crate::util::same_key(&self.last_key, key) {
             self.last_key.clear();
             self.last_key.extend_from_slice(key);
+            true
+        } else {
+            false
+        };
+
+        // If iterating in forward direction, then just checking the last key against
+        // current key would be sufficient.
+        if !self.opt.reverse && !key_changed {
+            self.table_iter.next();
+            return false;
         }
 
         loop {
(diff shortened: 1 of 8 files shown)


BusyJay:
Why not define `seek_to_first` and `seek_to_last` just like what rocksdb does?

GanZiheng:
> Why not define `seek_to_first` and `seek_to_last` just like what rocksdb does?

Currently, `rewind` is the same as `seek_to_first`, and `to_last` added int this pull request is acyually equivalent with `seek_to_last`. maybe I should rename these methods?

src/iterator.rs
```diff
@@ -274,6 +286,13 @@ impl<'a> Iterator<'a> {
 
     // Advances the iterator by one.
     pub fn next(&mut self) {
+        let key = BytesMut::from(&self.item().key.clone()[..]);
```
Connor1996, line 289: why clone?

src/iterator.rs
```diff
@@ -350,11 +397,92 @@ impl<'a> Iterator<'a> {
...
+            // Ignore the next candidate. Return the current one.
+            self.item = Some(item);
+            return true;
+        }
+    }
+
+    /// Just like `parse_item`, but used for `prev`.
+    ///
+    /// This function advances the iterator.
+    fn parse_item_for_prev(&mut self) -> bool {
```
Connor1996, line 414: can we reuse `parse_item` by passing a parameter indicating the direction?

src/iterator.rs
```diff
@@ -274,6 +286,13 @@ impl<'a> Iterator<'a> {
 
     // Advances the iterator by one.
     pub fn next(&mut self) {
+        let key = BytesMut::from(&self.item().key.clone()[..]);
+        let key_with_ts = key_with_ts(key, self.item().version);
+
+        // TODO: Consider add direction flag to avoid seek.
+        self.table_iter.seek(&key_with_ts);
```
Connor1996, line 293: why seek

src/iterator.rs
GanZiheng: Yes, we can combine `parse_item` and `parse_item_for_prev` once the logic is reviewed.

src/iterator.rs
GanZiheng: Since `parse_item` and `parse_item_for_prev` would advance and retreat the `table_iter` inside the `Iterator` respectively, what we should do in case of of `next` after `next` and `prev` after `next` are different. We may add flag or update `parse_item` and `parse_item_for_prev` to avoid seek.

src/iterator.rs
GanZiheng: Unnecessary, will fix this.

src/iterator.rs
GanZiheng: I added a flag `current_direction` to indicate the "direction" of the iterator. If the last operation is `prev`, it will be set to false, and we will advance the inner iterator double times in that case if we call `next` subsequently.

skyzh:
Generally LGTM. But I wonder if there is any use case inside AgateDB that we will need to do both forward iterating and backward iterating on a single iterator? If not, we can use const generics to implement forward/backward iterator using one piece of code. It will also simplifies the case on how to seek.

skyzh:
… some examples: https://github.com/singularity-data/risingwave/blob/main/src/storage/src/hummock/iterator/backward_concat.rs

GanZiheng:
> Generally LGTM. But I wonder if there is any use case inside AgateDB that we will need to do both forward iterating and backward iterating on a single iterator? If not, we can use const generics to implement forward/backward iterator using one piece of code. It will also simplifies the case on how to seek.

Previously, AgateDB has both forward and reverse iterators, but once the kind of iterator is determined, it could only go ahead but cannot go back. Current I support both forward and reverse iterator to go back.

skyzh:
> > Generally LGTM. But I wonder if there is any use case inside AgateDB that we will need to do both forward iterating and backward iterating on a single iterator? If not, we can use const generics to implement forward/backward iterator using one piece of code. It will also simplifies the case on how to seek.
> 
> Previously, AgateDB has both forward and reverse iterators, but once the kind of iterator is determined, it could only go ahead but cannot go back. Current I support both forward and reverse iterator to go back.

Yes, so I'm wondering whether this is necessary... I can hardly come up with a use case that users want to change direction on-the-fly.

GanZiheng:
> > > Generally LGTM. But I wonder if there is any use case inside AgateDB that we will need to do both forward iterating and backward iterating on a single iterator? If not, we can use const generics to implement forward/backward iterator using one piece of code. It will also simplifies the case on how to seek.
> > 
> > 
> > Previously, AgateDB has both forward and reverse iterators, but once the kind of iterator is determined, it could only go ahead but cannot go back. Current I support both forward and reverse iterator to go back.
> 
> Yes, so I'm wondering whether this is necessary... I can hardly come up with a use case that users want to change direction on-the-fly.

In fact, that's because RocksDB supports it and TiKV needs it :rofl:.

PR #188, opened 2022-08-23
Feature: Ingest External SST

wangnengjie:
Signed-off-by: wangnengjie [751614701@qq.com](mailto:751614701@qq.com)

See details in: https://github.com/tikv/agatedb/issues/187

### DIFF
--- a/proto/src/proto/meta.proto
+++ b/proto/src/proto/meta.proto
@@ -44,6 +44,7 @@ message ManifestChange {
   uint64 key_id  = 4;
   EncryptionAlgo encryption_algo = 5;
   uint32 compression = 6;   // Only used for CREATE Op.
+  uint64 global_version = 7; // Only used for file ingest, 0 means no global_version
 }
 
 message BlockOffset {
--- a/src/ingest.rs
+++ b/src/ingest.rs
@@ -284,7 +284,7 @@ mod tests {
     use super::IngestExternalFileOptions;
     use crate::{
         db::tests::run_agate_test, error::Result, key_with_ts, opt::build_table_options, Agate,
-        AgateOptions, Table, TableBuilder, Value,
+        AgateOptions, Table, TableBuilder, Value, IteratorOptions,
     };
 
     const BUILD_TABLE_VERSION: u64 = 10;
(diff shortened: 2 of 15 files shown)


wangnengjie:
The `next_txn_ts` won't update when reopening db? @GanZiheng

GanZiheng:
> 

Yes... I remember it is set to 1.

wangnengjie:
> Yes... I remember it is set to 1.

Well, that's... amazing... I debug a while for my failed test. I think I should make another pr to make it works right.

wangnengjie:
Generally done. You can have a look. @GanZiheng 

BTW, the `rustfmt.toml` use many flags that are only available in nightly channel while CI use stable channel. So it just failed though I run `make format` under nightly channel.

src/ops/oracle.rs
```diff
@@ -208,13 +224,39 @@ impl Oracle {
...
                     conflict_keys: txn.conflict_keys.clone(),
+                    conflict_ranges: vec![],
                 })
             }
 
             (ts, false)
         }
     }
 
+    pub(crate) fn new_ingest_commit_ts(
```
GanZiheng, line 235: Where is the conflict check?

proto/src/proto/meta.proto
```diff
@@ -44,6 +44,7 @@ message ManifestChange {
   uint64 key_id  = 4;
   EncryptionAlgo encryption_algo = 5;
   uint32 compression = 6;   // Only used for CREATE Op.
+  uint64 global_version = 7; // Only used for file ingest, 0 means no global_version
```
GanZiheng, line 47: Is it necessary to add `global_version` just for the ingest external file scenario?

src/ingest.rs
```diff
@@ -0,0 +1,806 @@
...
+pub enum PickLevelStrategy {
+    BaseLevel,
+    BottomLevel,
+}
+
+#[derive(Debug, Clone, Copy)]
+pub struct IngestExternalFileOptions {
+    pub commit_ts: u64,
+    pub move_files: bool,
+    pub failed_move_fall_back_to_copy: bool,
```
GanZiheng, line 22: Please add comment for this field.

src/ingest.rs
```diff
@@ -0,0 +1,806 @@
...
+            }
+        }
+        if self.overlap {
+            // files overlap with each other, assign commit_ts for each file
+            for file in self.files.iter_mut() {
+                let range = RangeInclusive::new(
+                    Bytes::copy_from_slice(user_key(file.table.as_ref().unwrap().smallest())),
+                    Bytes::copy_from_slice(user_key(file.table.as_ref().unwrap().biggest())),
+                );
+                let version = self.core.orc.new_ingest_commit_ts(vec![range], &self.opts);
```
GanZiheng, line 215: Can we only get commit timestamp once for all external files?

GanZiheng:
> Generally done. You can have a look. @GanZiheng
> 
> BTW, the `rustfmt.toml` use many flags that are only available in nightly channel while CI use stable channel. So it just failed though I run `make format` under nightly channel.

We currently use stable channel for the sake of avoiding CI failing too often. I think there is a format error in this version of code.

src/ops/oracle.rs
wangnengjie: https://github.com/wangnengjie/agatedb/blob/11161387d17c70bbe595afab67553ea1c9eef28e/src/ops/oracle.rs#L52

In `oracle::has_conflict`

proto/src/proto/meta.proto
wangnengjie: Yes. For ingested files, version of every key is invalid. Or if you have any better idea to assign version for keys in ingested files? Iter on every key and issue random writes is not a good idea.

src/ingest.rs
wangnengjie: I'd like to. But for files that overlap with each other, there might be same key. And if we get one commit_ts for all these files, there will be same key with same version in different SST which breaks the protection.

src/ops/oracle.rs
GanZiheng: > https://github.com/wangnengjie/agatedb/blob/11161387d17c70bbe595afab67553ea1c9eef28e/src/ops/oracle.rs#L52
> 
> In `oracle::has_conflict`

But it is not called in `new_ingest_commit_ts`.

wangnengjie:
> > Generally done. You can have a look. @GanZiheng
> > BTW, the `rustfmt.toml` use many flags that are only available in nightly channel while CI use stable channel. So it just failed though I run `make format` under nightly channel.
> 
> We currently use stable channel for the sake of avoiding CI failing too often. I think there is a format error in this version of code.

As what I have said, `rustfmt.toml` use many flags that are only available in nightly channel which makes the result of format different with stable channel. I will use stable channel to format the code later. So why not just remove all these flags that only available in nightly channel?

Does CI often failed in nightly channel? I saw that all tests (except the first one) run in nightly channel for the support of sanitizer...

proto/src/proto/meta.proto
GanZiheng: > Yes. For ingested files, version of every key is invalid. Or if you have any better idea to assign version for keys in ingested files? Iter on every key and issue random writes is not a good idea.

How about rewriting files to be ingested? In this way, we can also deal with the large value cannot be stored in value log.

src/ops/oracle.rs
wangnengjie: Oracle's conflict check is only for read keys of update transaction.

GanZiheng:
> 

Specifically, failed due to the clippy check. We can consider remove these flags or just turn back to nightly channel, I think that's not matter.

proto/src/proto/meta.proto
wangnengjie: I think it really costs to rewrite. The bandwidth on cloud device is precious. Large value can store to value log when compactioning ingested files (do we impl this?).

wangnengjie: > > Yes. For ingested files, version of every key is invalid. Or if you have any better idea to assign version for keys in ingested files? Iter on every key and issue random writes is not a good idea.
> 
> How about rewriting files to be ingested? In this way, we can also deal with the large value cannot be stored in value log.

I think it really costs to rewrite. The bandwidth on cloud device is precious. Large value can store to value log when compactioning ingested files (do we impl this?).

src/ops/oracle.rs
GanZiheng: > Oracle's conflict check is only for read keys of update transaction. And ingest task doesn't have read keys.

Yes, that's right.

proto/src/proto/meta.proto
GanZiheng: > 

Currently we only write value log when writing entry to LSM tree.

src/ops/oracle.rs
wangnengjie: > > Oracle's conflict check is only for read keys of update transaction. And ingest task doesn't have read keys.
> 
> Yes, that's right.

So we don't need to check conflict here during commiting ingest task~~

proto/src/proto/meta.proto
wangnengjie: > Currently we only write value log when writing entry to LSM tree.

Split during compaction is nice when the threshold changes dynamically or reopen db with different config.

Well, we can discuss this point tomorrow or offline.

wangnengjie: > > Yes. For ingested files, version of every key is invalid. Or if you have any better idea to assign version for keys in ingested files? Iter on every key and issue random writes is not a good idea.
> 
> How about rewriting files to be ingested? In this way, we can also deal with the large value cannot be stored in value log.

I think this will block read process. We need to get commit_ts (version) before rewrite and due to SSI protection, every read_ts after that should wait for this txn to finish (watermark).

PR #190, opened 2022-08-25
max_version support

wangnengjie:
Signed-off-by: wangnengjie <751614701@qq.com>

Support `max_version` and initialize `next_txn_ts` when open.

### DIFF
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -96,7 +96,10 @@ impl MemTable {
         self.skl.put(key, value);
 
         // update max version
-        inner.max_version = ts;
+        // Although we guarantee that the order of all updates pushed to write_ch is
+        // the same as getting commit_ts(version), user may set it non-monotonically
+        // in managed mode. So we should compare and update.
+        inner.max_version = inner.max_version.max(ts);
 
         Ok(())
     }
--- a/src/memtable.rs
+++ b/src/memtable.rs
@@ -275,6 +275,11 @@ mod tests {
                     .collect::<Vec<Arc<MemTable>>>(),
             ),
         };
+        assert_eq!(mem_tables.mutable.max_version(), 249);
+        assert_eq!(mem_tables.immutable[0].max_version(), 499);
+        assert_eq!(mem_tables.immutable[1].max_version(), 749);
+        assert_eq!(mem_tables.immutable[2].max_version(), 999);
+        assert_eq!(mem_tables.max_version(), 999);
         let view = mem_tables.view();
         for k in 0..4 {
             for i in k * 250..(k + 1) * 250 {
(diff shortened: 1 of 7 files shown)


wangnengjie:
> LGTM, how about adding unit tests for `MemTables`?

https://github.com/tikv/agatedb/blob/master/src/memtable.rs#L94-L100

seems that memtable now assumes version of adding keys in monotone increase mode. Is that as expected?

GanZiheng:
> > LGTM, how about adding unit tests for `MemTables`?
> 
> https://github.com/tikv/agatedb/blob/master/src/memtable.rs#L94-L100
> 
> seems that memtable now assumes version of adding keys in monotone increase mode. Is that as expected?

I think that is expected. Because we will call `memtable::MemTable::put` when writing updates to the LSM tree, and when using transaction with `write_ch_lock`, it is guaranteed that the timestamp of the updates pushed to the write channel is monotonically increasing.

But I think we may add comment for this or also update the `max_version` by comparing.

wangnengjie:
> > > LGTM, how about adding unit tests for `MemTables`?
> > 
> > https://github.com/tikv/agatedb/blob/master/src/memtable.rs#L94-L100
> > 
> > seems that memtable now assumes version of adding keys in monotone increase mode. Is that as expected?
> 
> I think that is expected. Because we will call `memtable::MemTable::put` when writing updates to the LSM tree, and when using transaction with `write_ch_lock`, it is guaranteed that the timestamp of the updates pushed to the write channel is monotonically increasing.
> 
> But I think we may add comment for this or also update the `max_version` by comparing.

For managed_db mode, commit_ts is set by user. It might not be guaranteed?

GanZiheng:
> > > > LGTM, how about adding unit tests for `MemTables`?
> > > 
> > > 
> > > https://github.com/tikv/agatedb/blob/master/src/memtable.rs#L94-L100
> > > seems that memtable now assumes version of adding keys in monotone increase mode. Is that as expected?
> > 
> > 
> > I think that is expected. Because we will call `memtable::MemTable::put` when writing updates to the LSM tree, and when using transaction with `write_ch_lock`, it is guaranteed that the timestamp of the updates pushed to the write channel is monotonically increasing.
> > But I think we may add comment for this or also update the `max_version` by comparing.
> 
> For managed_db mode, commit_ts is set by user. It might not be guaranteed?

Yes, so I think it would be better to update by comparing.

wangnengjie:
Done

src/memtable.rs
```diff
@@ -270,4 +294,54 @@ mod tests {
...
     }
+
+    #[test]
+    fn max_version() {
+        let skl = Skiplist::with_capacity(make_comparator(), 4 * 1024 * 1024, true);
+        let mem = MemTable::new(0, skl, None, AgateOptions::default());
+        for i in 200..300 {
+            let mut v = BytesMut::from(i.to_string().as_bytes());
+            append_ts(&mut v, i);
+            let v = v.freeze();
```
GanZiheng, line 305: We could replace this with `key_with_ts`.

GanZiheng (approved):
Rest LGTM.

GanZiheng:
PTAL @Connor1996.

GanZiheng (approved):

skyzh (approved):

BusyJay:
It will be helpful if the description of the PR or a linked issue can explain what is `max_version`.

wangnengjie:
> It will be helpful if the description of the PR or a linked issue can explain what is `max_version`.

Done. If somewhere is still puzzling, plz point out.

PR #196, opened 2024-04-13
*: update deps

BusyJay:
make github dependant bot happy.


skyzh (approved):

.github/workflows/test.yml
```diff
@@ -48,7 +48,9 @@ jobs:
       - name: Running Tests With Code Coverage 🚀
         run: cargo llvm-cov --all-features --workspace --lcov --output-path lcov.info
       - name: Upload Coverage To Codecov ⬆️
-        uses: codecov/codecov-action@v1
+        uses: codecov/codecov-action@v3
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
tabokie, line 53: Somehow this is overwriting raft-engine's coverage: https://app.codecov.io/gh/tikv/raft-engine

skyzh: I don't have admin access to the repo so someone else needs to generate a new token and fix that :)

BusyJay: The token is copied from the repo page, strange it will overwriting another repo 🤔

BusyJay: I setup tokens for both agatedb and raft-engine, and coverage works fine for agatedb now. Not sure if it's the cause, it's also possible that a wrong key is used in the first place.

PR #198, opened 2025-12-14
Fix build error on rust 1.9.20

stepinto:
Fix the belowing build error on rust 1.92.0.

  error: invalid format string: integer `1048576` does not fit into the
  type `u16` whose range is `0..=65535`
     --> src/table/tests.rs:353:29

### DIFF
--- a/src/table/tests.rs
+++ b/src/table/tests.rs
@@ -350,7 +350,9 @@ fn test_uni_iterator() {
 // TODO: concat iterators and merge iterators
 
 fn value(i: usize) -> Bytes {
-    Bytes::from(format!("{:01048576}", i)) // 1MB value
+    let mut result = i.to_string().into_bytes();
+    result.resize(1048576, 0);
+    result.into()
 }
 
 #[test]
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,7 +21,7 @@ jobs:
           access_token: ${{ github.token }}
       - uses: actions/checkout@v2
         name: Checkout 🛎️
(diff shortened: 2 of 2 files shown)


src/table/tests.rs
```diff
@@ -350,7 +350,11 @@ fn test_uni_iterator() {
 // TODO: concat iterators and merge iterators
 
 fn value(i: usize) -> Bytes {
-    Bytes::from(format!("{:01048576}", i)) // 1MB value
```
BusyJay, line 353: How about:
```
let result = i.to_string().into_bytes();
result.resize(1048576, 0);
```
?

stepinto: Good idea.  Fixed it using your approach.

skyzh:
please ping us for review once CI errors are fixed :) thank you

stepinto:
> please ping us for review once CI errors are fixed :) thank you

This seems to be an irrevelant error to this PR.  It is because this repo is using `actions/cache` v2, which has been deprecated since 2024-12.  Is there anyone working on fix it?

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/cache: v2`. Please update your workflow to use v3/v4 of actions/cache to avoid interruptions. Learn more: https://github.blog/changelog/2024-12-05-notice-of-upcoming-releases-and-breaking-changes-for-github-actions/#actions-cache-v1-v2-and-actions-toolkit-cache-package-closing-down
```

BusyJay:
Change "actions/cache@v2" to "actions/cache@v4" in .github/workflows/test.yml should work.

